PR #42, opened 2019-01-22
Fix possibly non existing parameter

yorodm:
Some LSP clients do not send `workspaceFolders` as a parameter and this line throws an AttributeError on those. Protocol allows for this field to be omitted

### DIFF
--- a/pygls/protocol.py
+++ b/pygls/protocol.py
@@ -510,8 +510,10 @@ class LanguageServerProtocol(JsonRPCProtocol, metaclass=LSPMeta):
         client_capabilities = params.capabilities
         root_uri = params.rootUri
         root_path = params.rootPath
-        workspace_folders = params.workspaceFolders or []
-
+        try:
+            workspace_folders = params.workspaceFolders or []
+        except AttributeError:
+            workspace_folders = []
         if root_uri is None:
             root_uri = from_fs_path(root_path) if root_path is not None else ''
 
--- a/pygls/protocol.py
+++ b/pygls/protocol.py
@@ -44,6 +44,7 @@ from .types import (ApplyWorkspaceEditParams, ApplyWorkspaceEditResponse,
                     ServerCapabilities, ShowMessageParams, WorkspaceEdit)
 from .uris import from_fs_path
 from .workspace import Workspace
+from .uris import to_fs_path
 
 logger = logging.getLogger(__name__)
 
@@ -509,23 +510,11 @@ class LanguageServerProtocol(JsonRPCProtocol, metaclass=LSPMeta):
 
         client_capabilities = params.capabilities
         root_uri = params.rootUri
-
-        try:
-            root_path = params.rootPath
-        except AttributeError:
-           root_path = None
-
-        try:
-            client_capabilities = params.capabilities
-        except AttributeError:
-            client_capabilities = None
-        try:
-            workspace_folders = params.workspaceFolders or []
-        except AttributeError:
-            workspace_folders = []
-        if root_uri is None:
-            root_uri = from_fs_path(root_path) if root_path is not None else ''
-
+        root_path = getattr(params, 'rootPath', None)
+        workspace_folders = getattr(params, 'workspaceFolders', [])
+        client_capabilities = getattr(params, 'capabilities', None)
+        if root_path is None:
+            root_path = to_fs_path(root_uri)
         self.workspace = Workspace(root_uri, self)
 
         for folder in workspace_folders:
--- a/pygls/protocol.py
+++ b/pygls/protocol.py
@@ -509,7 +509,16 @@ class LanguageServerProtocol(JsonRPCProtocol, metaclass=LSPMeta):
 
         client_capabilities = params.capabilities
         root_uri = params.rootUri
-        root_path = params.rootPath
+
+        try:
+            root_path = params.rootPath
+        except AttributeError:
+           root_path = None
+
+        try:
+            client_capabilities = params.capabilities
+        except AttributeError:
+            client_capabilities = None
         try:
             workspace_folders = params.workspaceFolders or []
         except AttributeError:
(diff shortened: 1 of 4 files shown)


augb:
[Here][1] is the relevant documentation, specifically, toward the end of the section:

>```/**
>  * The workspace folders configured in the client when the server starts.
>  * This property is only available if the client supports workspace folders.
> * It can be `null` if the client supports workspace folders but none are
> * configured.
>  *
>  * Since 3.6.0
>  */
> workspaceFolders?: WorkspaceFolder[] | null;
> ```

[1]: https://microsoft.github.io/language-server-protocol/specification#initialize

tests/test_language_server.py
```diff
@@ -48,6 +48,22 @@ def test_bf_initialize(client_server):
     assert hasattr(response, 'capabilities')
 
 
+def test_bf_initialize_spec(client_server):
+    client, _ = client_server
+    # Use a dictionary to send only the elements marked
+    # as required
+    response = client.lsp.send_request(
+        INITIALIZE,
+        {
```
augb: it might be clearer just to have lines 57-61 be an empty _dict_:
```
response = client.lsp.send_request(
    INITIALIZE,
    {}
).result(timeout=CALL_TIMEOUT)
```

augb:
One more thing, @yorodm. Please add your info to [CONTRIBUTORS.md](https://github.com/openlawlibrary/pygls/blob/master/CONTRIBUTORS.md), if you would like.

pygls/protocol.py
```diff
@@ -509,7 +509,16 @@ def bf_initialize(self, params: InitializeParams):
 
         client_capabilities = params.capabilities
         root_uri = params.rootUri
-        root_path = params.rootPath
+
```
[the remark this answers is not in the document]
yorodm: `rootUri` is mandatory so I believe server should fail if it's not provided (it should at least be null)

tests/test_language_server.py
yorodm: How do we test for paremeters required by the specification then?

pygls/protocol.py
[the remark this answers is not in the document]
augb: That is a good point, though, since the suggested test would have neither `rootUri` or `rootPath`. 

`rootUri` is derived from `rootPath` a little further below in the code:
```
if root_uri is None:
    root_uri = from_fs_path(root_path) if root_path is not None else ''
```
It may be that instead of setting `rootUri` to empty string, we should keep it as None, in this case. @danixeee Any thoughts?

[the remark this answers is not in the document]
danixeee: I would go with `getattr(params, 'workspaceFolders', [])`. For me it's more readable, especially if we will have few `try-except` blocks in a row.

PS: @yorodm thank you for your contribution.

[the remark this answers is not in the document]
yorodm: @danixee @augb Thanks to you guys for the framework. I'm building a Python LSP server with it and it's been wonderful to work with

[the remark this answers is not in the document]
augb: @yorodm Glad to hear it. When you are ready, please submit a PR to update [Implementations.md](https://github.com/openlawlibrary/pygls/blob/master/Implementations.md).

danixeee:
I have just tested the case when we are not providing the required fields during [Initialize request](https://microsoft.github.io/language-server-protocol/specification#initialize).

If we do something like this:

```
client_capabilities = getattr(params, 'capabilities', None)
root_uri = getattr(params, 'rootUri', None)
root_path = getattr(params, 'rootPath', None)
workspace_folders = getattr(params, 'workspaceFolders', [])
```

server will respond to the client, which, in my opinion, is not good, because other functionalities won't work properly due to the missing fields (if _capabilities_ or _rootUri_ is None). So, if capabilities are not provided, AttributeError will be raised and we will immediately see the cause of it.

We should get default values just for _optional_ fields in `InitalizeParams`, so just _root_path_ and _workspace_folders_ in our case.
 
There should be three test cases:

1. One of the _required_ fields is empty - Should raise AttributeError
2. All _optional_ fields are empty - Should work fine; default values should be checked
3. All fields are populated - Should work fine; values should be checked

@yorodm @augb What are your thoughts?

I can help with tests if you are not comfortable with pytest.

augb:
> ... server will respond to the client, which, in my opinion, is not good, because other functionalities won't work properly due to the missing fields (if capabilities or rootUri is None). So, if capabilities are not provided, AttributeError will be raised and we will immediately see the cause of it.

I agree. I think on the parameters that are truly required it would be better to raise an exception.

pygls/protocol.py
```diff
@@ -509,13 +509,12 @@ def bf_initialize(self, params: InitializeParams):
...
-        root_path = params.rootPath
-        workspace_folders = params.workspaceFolders or []
-
-        if root_uri is None:
-            root_uri = from_fs_path(root_path) if root_path is not None else ''
-
-        self.workspace = Workspace(root_uri, self)
+        root_path = getattr(params, 'rootPath', None)
+        workspace_folders = getattr(params, 'workspaceFolders', [])
+        client_capabilities = getattr(params, 'capabilities', None)
```
danixeee: This line should be removed (line 510. is fine).

yorodm: Oops, sorry, big mistake there.

augb: By removing line 514, line 510 should raise an attribute error if `capabilities` is `None`.

danixeee: Attribute error will be raised if request object does not have capabilities field.

@yorodm Good observation. This is from LSP specs:
```
/**
    * The rootUri of the workspace. Is null if no
    * folder is open. If both `rootPath` and `rootUri` are set
    * `rootUri` wins.
*/
```
So, _rootUri_ _can_ be _None_, but it's required, not sure if I got this.

I would say, just remove line 514, and if it's working on your end, we should be fine for now. The goal of this PR is to fix the bug that you detected, so you can move on with your project that is actually using `pygls`.

PR #76, opened 2019-08-16
Allow for a compiled executable to be set in VS Code extension

TiemenSch:

So I rewrote most of the VS Code extension example to a accommodate a slightly different setup. 
The detection of arguments to run the server with is now slightly different.

1. First and foremost, if a ``DebugTCP`` port is specified, it will attach to a running server on that port.
1. Then it looks for ``serverPath`` (executable) and ``serverArgs`` config settings, which allow for all kinds of different setups of the language server by just changing the VS Code user or workspace config. 
1. If no ``serverPath`` is specified, it will fallback to the ``python.pythonPath`` setting with the module as arguments.

I abstracted this so I could easily switch to a compiled version of the language server. To document that, I added a small section to the ``advanced_usage`` docs that describes how to compile the server and switch to it using the newly added settings.


danixeee:
This looks good, thank you for the PR.
I suggest extending the existing _JSON language server_ example, since most of the files are not changed. We should modify just `extension.ts` to allow starting the compiled server. If you can work on that and I will take a look and try it out when it's ready.

examples/json-extension/client/src/extension.ts
```diff
@@ -20,80 +20,92 @@
...
 
 export function activate(context: ExtensionContext) {
-  if (isStartedInDebugMode()) {
-    // Development - Run the server manually
-    client = startLangServerTCP(2087);
-  } else {
-    // Production - Client is going to run the server (for use within `.vsix` package)
-    const cwd = path.join(__dirname, "../");
-    const pythonPath = workspace.getConfiguration("python").get<string>("pythonPath");
+    const debugPort = process.env.DebugTCP;
```
danixeee: Function `isStartedInDebugMode` is not used anymore. I think we should keep port hardcoded and change it in another PR, because we should read that environment variable when starting the server from VS Code.

examples/json-extension/client/src/extension.ts
```diff
@@ -20,80 +20,92 @@
...
+    const config = getConfig("jsonServer");
+    if (debugPort) {
+        // Debug port: server should be running externally on specified port.
+        client = startLangServerTCP(Number(debugPort));
+    } else {
```
danixeee: I think it would be easier for user to try it out if you add `serverPath` and `serverArgs` to environment variables [here](https://github.com/openlawlibrary/pygls/blob/master/examples/json-extension/.vscode/launch.json#L18) and push that file.
Then, instead doing `config.get<string>("serverPath");` you can get config values with `process.env.SERVER_EXECUTABLE_PATH` and `process.env.SERVER_ARGS`.

Finally, we should update docs and say that [this](https://github.com/openlawlibrary/pygls/blob/master/examples/json-extension/.vscode/launch.json#L19) environment variable needs to be _False_ in order to run pre-compiled server.

```typescript
if (isStartedInDebugMode()) {
  // Development - Run the server manually
  client = startLangServerTCP(2087);
} else {
  // Client is going to run the server
  const cwd = path.join(__dirname, "../");

  const serverPath = process.env.SERVER_EXECUTABLE_PATH;
  const serverArgs = process.env.SERVER_ARGS;

  if (serverPath) {
    // Run pre-compiled server
    client = startLangServer(serverPath, serverArgs, cwd);
  } else {
    // Run server from source
    client = startLangServer(pythonPath, ["-m", "server"], cwd);
  }
}
```

@TiemenSch What are your thoughts here?

TiemenSch: Personally, I really like the VS Code settings approach because this makes it easier for users of the extension to point it to any executable/args combination even when not debugging. These settings are also neatly documented out of the box in VS Code's settings menu's.

 It also doesn't require you to clog the environment variables any more than necessary :).

docs/source/pages/advanced_usage.rst
```diff
@@ -456,3 +456,31 @@ Workspace methods that can be used for user defined features are:
     def apply_edit(self, edit: WorkspaceEdit, label: str = None) -> ApplyWorkspaceEditResponse:
         # Omitted
 
+Compilation
```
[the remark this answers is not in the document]
TiemenSch: Ah, that would probably require an install of the package, too. I could also merge the files into a single file called ``json_server.py`` or something similar. Or would you like to stick with the package layout?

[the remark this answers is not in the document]
danixeee: I am fine with a script, you can just copy a content from `__main__.py` then and fix imports in tests if needed.

examples/json-extension/client/src/extension.ts
danixeee: That's fine, I agree mostly, but I am usually running the example without opening any folder and I don't want to add those settings under user-settings.

You can set those variables depending on OS:

```json
...
 "env": {
                "VSCODE_DEBUG_MODE": "true"
            },
            "osx": {
                "env": {
                    "SERVER_EXECUTABLE_PATH": "build/json_server.dist/json_server",
                }
            },
            "linux": {
                "env": {
                    "SERVER_EXECUTABLE_PATH": "build/json_server.dist/json_server",
                }
            },
            "windows": {
                "env": {
                    "SERVER_EXECUTABLE_PATH": "build/json_server.dist/json_server.exe",
                }
            },
...
```

and append them on `cwd` path. This is just an example and for production, someone will need to figure out how to package and pre-compiled server, depending on platform.

That's just from my perspective when trying to run example and review a PR. :)

TiemenSch: Ah, we could look for both env variables and settings (in what order?) to account for this.

I just open a folder with the right workspacesettings depending on what use-case I want to debug ;). It's not that I don't trust myself changing my system's path, but rather that I would not want end-users to have to do so.

danixeee: You could check settings primary and then env vars, but I guess you won't need to do this.
In case that I am suggesting, you (and any other user) won't need to change and nor setup anything. https://github.com/openlawlibrary/pygls/pull/76#discussion_r327674536 is a content of `launch.json` file under `json-example/.vsode` directory and has relative path to the _json-example_ directory. The only thing that they need to do is to install and compile server and then launch the client.

danixeee:
Process of creating an executable for a project based on _pygls_ should be the same as for any other python project, so I am closing this.

PR #78, opened 2019-08-21
Fix full updates

pappasam:
## Description

FULL syncs were broken for me unless I set the server to only accept FULL syncs. This PR fixes the bug.

## Discussion

Even if a server accepts INCREMENTAL SyncKinds, clients may request a FULL SyncKind. In LSP 3.x, clients make this request by omitting both Range and RangeLength from their request. Consequently, the attributes "range" and "range_length" will be missing from FULL content update client requests in the pygls Python library.

The client library I use is [LanguageClient-neovim](https://github.com/autozimu/LanguageClient-neovim), which only supports full updates at present. It explicitly removes the attributes Range and RangeLength from its request when they are None, as the LSP specification requests. This clearly flows to the lower-level `_apply_incremental_change(` function.

I haven't been able to fully wrap my head around the lower-level parts of this library (types.py -> protocol.py), so maybe you can find a more elegant solution than this one. That said, this is currently a breaking bug, and this PR fixes the problem, enabling me to not explicitly override the server's sync settings to only support FULL.

NOTE: I haven't updated tests (not sure which to update) or the ChangeLog (will wait until I get confirmation that you like this PR before doing that).

Docs:

* DidChange: https://microsoft.github.io/language-server-protocol/specification#textDocument_didChange
* Optional Properties: https://www.typescriptlang.org/docs/handbook/interfaces.html#optional-properties

### DIFF
--- a/pygls/workspace.py
+++ b/pygls/workspace.py
@@ -110,7 +110,7 @@ class Document(object):
             Even if a server accepts INCREMENTAL SyncKinds, clients may request
             a FULL SyncKind. In LSP 3.x, clients make this request by omitting
             both Range and RangeLength from their request. Consequently, the
-            attributes "range" and "range_length" will be missing from FULL
+            attributes "range" and "rangeLength" will be missing from FULL
             content update client requests in the pygls Python library.
 
         Improvements:
@@ -124,7 +124,7 @@ class Document(object):
         """
         if (
             hasattr(change, 'range') and
-            hasattr(change, 'range_length') and
+            hasattr(change, 'rangeLength') and
             self._is_sync_kind_incremental
         ):
             self._apply_incremental_change(change)
@@ -132,7 +132,7 @@ class Document(object):
             self._apply_none_change(change)
         elif not (
             hasattr(change, 'range') or
-            hasattr(change, 'range_length')
+            hasattr(change, 'rangeLength')
         ):
             self._apply_full_change(change)
         else:
--- a/pygls/workspace.py
+++ b/pygls/workspace.py
@@ -106,9 +106,6 @@ class Document(object):
         Performs either INCREMENTAL, FULL, or NONE synchronization based on
         both the Client request and server capabilities.
 
-        Raises ValueError if the client provides an ambiguous or unsupported
-        request.
-
         INCREMENTAL versus FULL synchronization:
             Even if a server accepts INCREMENTAL SyncKinds, clients may request
             a FULL SyncKind. In LSP 3.x, clients make this request by omitting
--- a/pygls/workspace.py
+++ b/pygls/workspace.py
@@ -45,32 +45,15 @@ class Document(object):
         self._local = local
         self._source = source
 
-        self._sync_func = {
-            TextDocumentSyncKind.FULL: self._apply_full_change,
-            TextDocumentSyncKind.INCREMENTAL: self._apply_incremental_change
-        }.get(sync_kind, None)
+        self._is_sync_kind_full = sync_kind == TextDocumentSyncKind.FULL
+        self._is_sync_kind_incremental = sync_kind == TextDocumentSyncKind.INCREMENTAL
+        self._is_sync_kind_none = sync_kind == TextDocumentSyncKind.NONE
 
     def __str__(self):
         return str(self.uri)
 
     def _apply_incremental_change(self, change: TextDocumentContentChangeEvent) -> None:
-        """Apply an either INCREMENTAL or FULL text change to the document
-
-        Note on INCREMENTAL versus FULL:
-            Even if a server accepts INCREMENTAL SyncKinds, clients may request
-            a FULL SyncKind. In LSP 3.x, clients make this request by omitting
-            both Range and RangeLength from their request. Consequently, the
-            attributes "range" and "range_length" will be missing from FULL
-            content update client requests in the pygls Python library.
-        """
-        if (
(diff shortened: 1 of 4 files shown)


pygls/workspace.py
```diff
@@ -54,15 +54,26 @@ def __str__(self):
...
+        """Apply an either INCREMENTAL or FULL text change to the document
+
+        Note on INCREMENTAL versus FULL:
+            Even if a server accepts INCREMENTAL SyncKinds, clients may request
+            a FULL SyncKind. In LSP 3.x, clients make this request by omitting
+            both Range and RangeLength from their request. Consequently, the
+            attributes "range" and "range_length" will be missing from FULL
+            content update client requests in the pygls Python library.
+        """
+        if (
```
danixeee: I didn't know that, but if that's the case, I think we should keep logic on which text document synchronization method to keep in `apply_change` method.

```python
def apply_change(self, change: TextDocumentContentChangeEvent) -> None:
    """Apply change to a document, depending on synchronization kind."""
    client_sync_kind = TextDocumentSyncKind.FULL if not (hasattr(change, 'range') or hasattr(change, 'range_length')) else TextDocumentSyncKind.INCREMENTAL
    if self._sync_func is None:
        self._sync_func = {
            TextDocumentSyncKind.FULL: self._apply_full_change,
            TextDocumentSyncKind.INCREMENTAL: self._apply_incremental_change
        }.get(min(client_sync_kind, self._server_sync_kind), None)
        
    try:
        self._sync_func(change)
    except TypeError:
        # Don't sync document if sync kind is TextDocumentSyncKind.NONE
        pass 
```

And you will have to slightly modify `__init__` method. What do you think?

pappasam: Great suggestion, I'll work on this tomorrow afternoon and send an update then!

pappasam: @danixeee I've made your suggested updates (with some minor changes). I've put lots of information in the docstrings / comments / commit message, but please let me know if you have any questions!

pappasam:
@danixeee I've made your suggested updates (with some minor changes). I've put lots of information in the docstrings / comments / commit message, but please let me know if you have any questions!

pygls/workspace.py
```diff
@@ -97,16 +91,63 @@ def _apply_incremental_change(self, change: TextDocumentContentChangeEvent) -> N
...
-            self._sync_func(change)
-        except TypeError:
-            # Don't sync document if sync kind is TextDocumentSyncKind.NONE
-            pass
+        """Apply a text change to a document, considering TextDocumentSyncKind
+
+        Performs either INCREMENTAL, FULL, or NONE synchronization based on
+        both the Client request and server capabilities.
+
+        Raises ValueError if the client provides an ambiguous or unsupported
```
danixeee: `ValueError` is not raised in this method, please remove this part of _docstring_.

pappasam:
NOTE: the following comment discusses next steps AFTER this PR is merged. This PR fixes a bug, while this comment discusses a design overhaul.

I strongly prefer the suggestion to "keep mappings between LSP methods and their parameters types and on each request to construct the exact object for invoked method." It could be a decent amount of maintenance, but it would also be a great way to explicitly document how LSP directly translates to pygls.

Also, regarding nested complex types, this problem is easily solved using a library like https://github.com/Dobiasd/undictify (one of my favorites for safe, recursive JSON deserialization). Only problem: this library supports Python 3.6+, while pygls supports python 3.5. Popular libraries like https://pypi.org/project/black/ only support 3.6+ because of the additional features / guarantees provided in that version: https://docs.python.org/3.6/whatsnew/3.6.html#summary-release-highlights.

If we choose to support 3.5+, we can consider libraries like https://simplejson.readthedocs.io/en/latest/, which might help us parse nested classes.

Please let me know what our constraints are:

1. Can we, in theory, drop support for 3.5?
2. If no, is is acceptable to add a 3rd party dependency to this library? If no, I can try my hand at writing a recursive JSON -> nested namedtuple decoder.

danixeee:
@pappasam Could you please update the _CHANGELOG_ so I can approve it?

pappasam:
@danixeee done

pygls/workspace.py
```diff
@@ -97,16 +91,60 @@ def _apply_incremental_change(self, change: TextDocumentContentChangeEvent) -> N
...
+            and all other LSP primitive types, to set "Optional" interface
+            attributes from the client to "None". The "hasattr" check is
+            admittedly quite ugly; while it is appropriate given our current
+            state, there are plenty of improvements to be made. A good place to
+            start: require more rigorous de-serialization efforts when reading
+            client requests in protocol.py.
+        """
+        if (
+            hasattr(change, 'range') and
+            hasattr(change, 'range_length') and
```
danixeee: Attribute names passed from the client are in camelCase format.

`range_length` -> `rangeLength`

pygls/workspace.py
```diff
@@ -97,16 +91,60 @@ def _apply_incremental_change(self, change: TextDocumentContentChangeEvent) -> N
...
+            self._is_sync_kind_incremental
+        ):
+            self._apply_incremental_change(change)
+        elif self._is_sync_kind_none:
+            self._apply_none_change(change)
+        elif not (
+            hasattr(change, 'range') or
+            hasattr(change, 'range_length')
```
danixeee: `range_length` -> `rangeLength`

pappasam:
Great catch!

PR #90, opened 2020-01-06
types: add __repr__ and __eq__ support to Range and Location

perrinjerome:
## Description 

This helps debugging and simplify testing, where we can just assert that ranges or locations are equals.

## Code review checklist (for code reviewer to complete)

- [x] Pull request represents a single change (i.e. not fixing disparate/unrelated things in a single PR)
- [x] Title summarizes what is changing
- [x] Commit messages are meaningful (see [this][commit messages] for details)
- [x] Tests have been included and/or updated, as appropriate
- [ ] Docstrings have been included and/or updated, as appropriate
- [ ] Standalone docs have been updated accordingly
- [x] [CONTRIBUTORS.md][contributors] was updated, as appropriate
- [x] Changelog has been updated, as needed (see [CHANGELOG.md][changelog])

[changelog]: https://github.com/openlawlibrary/pygls/blob/master/CHANGELOG.md
[commit messages]: https://chris.beams.io/posts/git-commit/
[contributors]: https://github.com/openlawlibrary/pygls/blob/master/CONTRIBUTORS.md

Issue #81, opened 2019-08-29
LSP method-parameter mappings


danixeee:
Initial discussion started in #78.

Currently, data received by the client is deserialized to `dict` object which is then converted to `namedtuple` object. Reason behind this is that we want to access fields with dot notation.
Some clients are not serializing/passing _null_ fields, which throws `AttributeError` when trying to access them in server methods.

Here are couple of ideas/solutions which need further discussion:

- https://github.com/openlawlibrary/pygls/pull/78#pullrequestreview-281366895
- https://github.com/openlawlibrary/pygls/pull/78#issuecomment-526247971

danixeee:
Another idea is is to override `__getattribute__` method of `namedtuple` instances, which will allow us to use dot notation for fields that are not passed. Then we can access those fields and get _None_ instead of _AttributeError_:

```python
def _safe_getattr(self, name):
    try:
        return object.__getattribute__(self, name)
    except AttributeError:
        return None

def safenamedtuple(cls, *args, **kwargs):
    obj = namedtuple(cls, *args, **kwargs)
    obj.__getattribute__ = _safe_getattr
    return obj
```

pappasam:
@danixeee thanks for kicking this off!

My 2-cents: if we rely on less Python magic, this Language Server will serve as excellent documentation for the LSP itself, which will likely increase the number of Python-language LSP implementations that are created using pygls. Therefore, I'm a big fan of explicitly mapping map pre-defined Python typed NamedTuples to the expected messages from the client. That way, we can handle errors clearly and immediately for poorly-configured clients.

danixeee:
@pappasam I partially agree with you. I've added type annotations to most of the parameters so we know which objects are expect in each _LSP_ method and to have better IDE support while developing. I think we won't benefit much with explicit type mappings, without runtime type checking.

I am happy to discuss this further and figure out the appropriate solution.

pappasam:
@danixeee There is a performance argument, and a readability argument, in favor of explicit mappings type mappings. I also think I've found a way to do it elegantly, without needing any 3rd party dependencies, and without doing any runtime validation. All we'll do is painlessly associate an LSP method name, as we expect to receive from the client, to each of our classes. At the time of this association, we assemble the classes into a dictionary once (at server initialization time) that can easily be used for method lookups.

The following sample code:

* does not have runtime validation
*  will ensure that the values we work with have the appropriate attributes defined on them
* is pretty DRY

```python
[...]

PR #92, opened 2020-01-06
types: use enum class for enums

perrinjerome:
## Description 

Using enums allows more precise type checking. The current code in types.py was already using *Kind classes as if they were enums, which mypy reported as type errors:
    
    pygls/types.py:377: error: Incompatible default for argument "severity" (default has type "int", argument has type "DiagnosticSeverity")
    pygls/types.py:663: error: Incompatible default for argument "trace" (default has type "str", argument has type "Optional[Trace]")

WatchKind is handled a bit differently, because valid values can be combined with |, like the default value 7.

### DIFF
--- a/pygls/protocol.py
+++ b/pygls/protocol.py
@@ -15,6 +15,7 @@
 # limitations under the License.                                           #
 ############################################################################
 import asyncio
+import enum
 import functools
 import json
 import logging
@@ -67,6 +68,13 @@ def call_user_feature(base_func, method_name):
         return ret_val
 
     return decorator
+
+
+def default_serializer(o):
+    """JSON serializer for complex objects."""
+    if isinstance(o, enum.Enum):
+        return o.value
+    return o.__dict__
 
 
 def deserialize_message(data):
@@ -370,7 +378,7 @@ class JsonRPCProtocol(asyncio.Protocol):
             return
 
         try:
-            body = json.dumps(data, default=lambda o: o.__dict__)
+            body = json.dumps(data, default=default_serializer)
             content_length = len(body.encode(self.CHARSET)) if body else 0
 
             response = (
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,10 @@ and this project adheres to [Semantic Versioning][semver].
 ### Added
 
 - Add comparisons and repr support to Range and Location types ([#90])
+- Use python enums in types module ([#92])
 
 [#90]: https://github.com/openlawlibrary/pygls/pull/90
+[#92]: https://github.com/openlawlibrary/pygls/pull/92
 
 ## [0.8.1] - 09/05/2019
 
--- a/pygls/types.py
+++ b/pygls/types.py
@@ -29,6 +29,8 @@ https://microsoft.github.io/language-server-protocol/specification#client_unregi
 """
 
 
+import enum
+import sys
 from typing import Any, Callable, Dict, List, Optional, Union
 
 from .features import (CODE_ACTION, CODE_LENS, CODE_LENS_RESOLVE, COMPLETION,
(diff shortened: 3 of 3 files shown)


pygls/protocol.py
```diff
@@ -369,8 +370,12 @@ def _send_data(self, data):
         if not data:
             return
 
+        def default(o):
```
danixeee: Could you just move this function under `call_user_feature`(line 70.) and perhaps rename it to `default_serializer` or something similar?

perrinjerome: Thanks, good idea, I pushed more fixup commits to address this.

danixeee:
@perrinjerome PR https://github.com/openlawlibrary/pygls/pull/90 is merged. If you feel comfortable, you can rebase branch with master and update the changelog for this PR.

perrinjerome:
Thanks @danixeee . I was not familiar with `enum` module, and after doing a bit more research, I realised that using `enum.Enum` was incorrect, because protocol is not deserialised as actual objects from `pygls.types` module.

For example when language server is implemented as:

```python
@server.feature(COMPLETION, trigger_characters=[','])
def completions(params: CompletionParams = None) -> CompletionList:
    """Returns completion items."""
    return CompletionList(False, [])
```

what this function receive as `params` argument is not a `CompletionParams`, but a "simple" `NamedTuple`, as protocol messages are deserialized here:

https://github.com/openlawlibrary/pygls/blob/df8950dac6539e387817d3f310c18465d8997204/pygls/protocol.py#L83

This also means that `params.context.triggerKind`, which is supposed to be of type `TriggerKind` enum is not an enum but a plain `int` and tests such as ( this comes from [cmake-language-server](https://github.com/regen100/cmake-language-server/blob/66af586b2aa4da1a21b54e566f81d250acb0a848/src/cmake_language_server/server.py#L43-L46) )

```python
        @self.feature(COMPLETION, trigger_characters=['{', '('])
        def completions(params: CompletionParams):
            if (params.context.triggerKind ==
                    CompletionTriggerKind.TriggerCharacter):
```
will be always false, because int are different from enum.

This small issue also interesting in context of https://github.com/openlawlibrary/pygls/issues/81 


But the problem seems that I wrongly used `enum.Enum` in this patch. When using `enum.IntEnum`, everything works fine  ( reference doc: https://docs.python.org/3/library/enum.html#intenum ) :

```
>>> import enum
>>> class wrong(enum.Enum):
...   one = 1
... 
>>> class ok(enum.IntEnum):
...   one = 1
... 
>>> 1 == wrong.one
False
>>> 1 == ok.one
True
```

So I will amend this patch to use `IntEnum` for int.

I will also change the string enums to inherit from `str` as suggested in https://docs.python.org/3/library/enum.html#others 

I will also revisit the `WatchKind` case to use an [`enum.IntFlag`]( https://docs.python.org/3/library/enum.html#intflag).

perrinjerome:
> I will also revisit the `WatchKind` case to use an [`enum.IntFlag`](https://docs.python.org/3/library/enum.html#intflag).

That's what I did as part of https://github.com/openlawlibrary/pygls/pull/92/commits/c0623414c9e2e956298dc6f2568ef3da042f5977  but `IntFlag` is *New in version 3.6.*, so that's was not OK.
I'll revisit this, hopefully later this week.

perrinjerome:
I amended to support python 3.5 , the change compared to previous version was https://github.com/openlawlibrary/pygls/commit/354ff54e4edbe835b5dc3bcd328fd8500fce1526 .

Then I rebased on master, squashed all the intermediate "fixup" commits from this branch, added an entry in change log, edited commit message to explain the need for `enum.IntEnum` and `str` subclassing and push-force the result in this branch.

Please take another look, I think this is OK now.

perrinjerome:
Thank you @danixeee . I also learned about enums in the process :) Don't worry about the delay, everybody is busy.

PR #117, opened 2020-05-02
Fix utf-16 to utf-32 character offsets.

### LINKED ISSUE #116: Make sure you handle UTF-16 character offsets
One of the major downsides of pyls is that it craps out at UTF-16 offsets.

I just want to make sure that this library doesn't fail in the same way. Just to be clear, the LSP spec says:

> A position inside a document (see Position definition below) is expressed as a zero-based line and character offset. **The offsets are based on a UTF-16 string representation**. So a string of the form a𐐀b the character offset of the character a is 0, the character offset of 𐐀 is 1 and the character offset of b is 3 since 𐐀 is represented using two code units in UTF-16.

Since python internally works with UTF-32 and indexes strings with UTF-32, you have to be very careful in translating LSP-ranges to pygls ranges. It is not as simple as converting a (row, col) to a character offset.

If you already accounted for this, then you may close this issue.

deathaxe:
Closes #116.

Python uses utf-32 for strings, but the language-server-protocol uses utf-16 to encode strings. Hence higher level unicode code points are represented by two utf-16 code units. Thus the `position.character` needs to be adjusted in case of the presence of such unicode characters.

see: https://github.com/openlawlibrary/pygls/issues/116, https://github.com/microsoft/language-server-protocol/issues/376

This PR therefore introduces two methods to the `Documents` class to handle conversion between utf-16 and utf-32 column offsets for general use.

The new methods are used in the `offset_at_position()`, `word_at_position()` and `_apply_incremental_change()` methods.

Notes:

1. The conversion needs information about the line's content. Hence it feels natural to add the functionality to the Document.

2. This change is unfortunately not transparent. Means a real server needs to proxy all positions through

   a) `Document.rowcol_to_position()` in order to create a `Position` with correct utf-16 code units to be sent to the client.

   b) `Document.position_to_rowcol()` to receive a corrected (row, col) tuple.

3. The methods convert between tuple <-> Position to make the difference clear and because row and column are needed explicitly in the current use cases anyway. A new type could be invented but it would need to make the difference obvious.

5. The `''.join(..)` statement in `offset_at_position()` is replaced as creating a maybe huge string just to count characters while the length of each line is a stored attribute seems too expensive.

_Please replace this description with a concise description of this Pull Request._

### DIFF
--- a/pygls/workspace.py
+++ b/pygls/workspace.py
@@ -57,10 +57,8 @@ class Document(object):
         text = change.text
         change_range = change.range
 
-        start_line = change_range.start.line
-        start_col = change_range.start.character
-        end_line = change_range.end.line
-        end_col = change_range.end.character
+        start_line, start_col = self.position_to_rowcol(change_range.start)
+        end_line, end_col = self.position_to_rowcol(change_range.end)
 
         # Check for an edit occuring at the very end of the file
         if start_line == len(self.lines):
@@ -150,26 +148,60 @@ class Document(object):
     def lines(self) -> List[str]:
         return self.source.splitlines(True)
 
-    def offset_at_position(self, position: Position) -> int:
-        """Return the character offset pointed at by the given position.
+    def position_to_rowcol(self, position: Position) -> tuple:
+        """Convert a LSP position into a row, column pair.
 
-        A position inside a document is expressed as a zero-based line and
-        character offset. The offsets are based on a UTF-16 string representation.
-
-        This method converts UTF-16 character offsets to UTF-32.
+        This method converts the Position's character offset
+        from UTF-16 code units to UTF-32 code points.
 
         The offset of the closing quotation mark in x="😋" is
         - 5 in UTF-16 representation
         - 4 in UTF-32 representation
-        """
-        column = 0
-        if position.line < len(self.lines):
-            column = position.character
-            for ch in self.lines[position.line][:position.character]:
-                if ord(ch) > 0xFFFF:
-                    column -= 1
 
-        return column + sum(len(line) for line in self.lines[:position.line])
+        A python application can't use the character memeber of `Position`
+        directly as per specification it is represented as a zero-based line and
+        character offset based based on a UTF-16 string representation.
+
+        All characters whose codepoint exeeds the Basic Multilingual Plane are
+        represented by 2 UTF-16 code units.
+
+        see: https://github.com/microsoft/language-server-protocol/issues/376
+        """
+        row = len(self.lines)
+        col = 0
+        if row > position.line:
+            row = position.line
+            col = position.character
+            for ch in self.lines[row][:position.character]:
+                if ord(ch) > 0xFFFF:
+                    col -= 1
+        return (row, col)
+
+    def rowcol_to_position(self, row: int, col: int) -> Position:
+        """Convert a row, column pair into a LSP Position.
+
+        This method converts the `col` argument from UTF-32 code points to
+        to UTF-16 code units and returns a `Position` object.
+
+        A python application can't use the character memeber of `Position`
+        directly as per specification it is represented as a zero-based line and
+        character offset based based on a UTF-16 string representation.
+
+        All characters whose codepoint exeeds the Basic Multilingual Plane are
+        represented by 2 UTF-16 code units.
+        """
+        line = len(self.lines)
+        character = 0
+        if line > row:
+            line = row
+            character = sum(1 + int(ord(ch) > 0xFFFF) for ch in self.lines[line][:col])
+
+        return Position(line, character)
+
+    def offset_at_position(self, position: Position) -> int:
+        """Return the character offset pointed at by the given position."""
+        row, col = self.position_to_rowcol(position)
+        return col + sum(len(line) for line in self.lines[:row])
 
     @property
     def source(self) -> str:
@@ -185,14 +217,11 @@ class Document(object):
         if position.line >= len(self.lines):
             return ''
 
-        line = self.lines[position.line]
-        i = position.character
-        for ch in line[:position.character]:
-            if ord(ch) > 0xFFFF:
-                i -= 1
+        row, col = self.position_to_rowcol(position)
+        line = self.lines[row]
         # Split word in two
-        start = line[:i]
-        end = line[i:]
+        start = line[:col]
+        end = line[col:]
 
         # Take end of start and start of end to find word
         # These are guaranteed to match, even if they match the empty string
--- a/pygls/workspace.py
+++ b/pygls/workspace.py
@@ -76,8 +76,13 @@ def position_from_utf16(lines: List[str], position: Position) -> Position:
     Returns:
         The position with `character` being converted to utf-32 code units.
     """
-    position.character -= utf16_unit_offset(lines[position.line][:position.character])
-    return position
+    try:
+        return Position(
+            position.line,
+            position.character - utf16_unit_offset(lines[position.line][:position.character])
+        )
+    except IndexError:
+        return Position(len(lines), 0)
 
 
 def position_to_utf16(lines: List[str], position: Position) -> Position:
@@ -105,8 +110,13 @@ def position_to_utf16(lines: List[str], position: Position) -> Position:
     Returns:
         The position with `character` being converted to utf-16 code units.
     """
-    position.character += utf16_unit_offset(lines[position.line][:position.character])
-    return position
+    try:
+        return Position(
+            position.line,
+            position.character + utf16_unit_offset(lines[position.line][:position.character])
+        )
+    except IndexError:
+        return Position(len(lines), 0)
 
 
 def range_from_utf16(lines: List[str], range: Range) -> Range:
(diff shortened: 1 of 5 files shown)


rwols:
The problems don't end here. For instance, the consumers of this library receive raw LSP positions:

https://github.com/muffinmad/anakin-language-server/blob/65a23097277ee7999f4ae824f11f034921121b64/anakinls/server.py#L226-L229

This means consumers are receiving LSP-space positions, and are assuming they're getting "Python" / "Jedi" space completions.

danixeee:
This code was initially copied from [pyls](https://github.com/palantir/python-language-server) and I wasn't aware of this. Thanks for these fixes. Could you please update the changelog and you can also add yourself to contributors doc if you wish.

deathaxe:
They are not aware of it, too. To be honest, I was trying to find an approach which transparently converts those column offsets in order to avoid bothering server developers with that detail.

Do you see any chance to do something like that?

One further question arose yesterday. What If a server needs to to conversion after an expensive job, which was computed in a thread and the document was modified in the meanwhile?

I have a feeling of my solution to not be the best choise then, but I am not too sure about it. Maybe should add global functions or make them a method of Position and passing a `string` which is used to compute the column offsets.

Maybe we can workout a proper pythonic solution? 

You may read that PR as an initial point of discussion.

danixeee:
@deathaxe I will test it again, but it seemed to be working fine.

> What If a server needs to to conversion after an expensive job, which was computed in a thread and the document was modified in the meanwhile?

I am not sure how we will solve this, but it doesn't have to be in this PR. Do you have an example when this could happen?

It makes sense to me to have methods / global functions for this conversion which will be called on each request whose parameter extends [TextDocumentPositionParams](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams) and possibly some other like _DidChangeTextDocumentParams_. That way other developers won't have to bother with this, as you said. If you wish, you can give it a try in this PR, otherwise, I think this is fine enough for now.

What do you think?

deathaxe:
The code of this PR makes sure to correctly sync the document incrementally, which might be one of the most importent things to start with. Hence I'd be fine to merge it without further changes to keep the scope limited to this one aspect.

Just want to make sure not to add some unluckily choosen methods which might cause trouble in the long term as I just started to play with this library last weekend.

> Do you have an example when this could happen?

Imagine the `_valideate_json()` in your example server was a time expensive asynchronous task which is executed in a separate thread. A language server could have received another didChange notification befor it finishes which updates the document by adding unicode characters to a line which would also be part of the `_validate_json()` command's results.

The `Position` or `Range` parts of the `Diagnostic` being sent to the client needs to be adjusted to utf-16 code units as well. As the document changed the conversion would run against the wrong piece of content.

Ok, we can argue the Diagnostics not to be valid in that situation anymore anyway and drop them, but I am not sure whether there are situations in which that could be important to run against an up to date string.

I thought about that possibility theroretically when diving into anakin-languages-server.

deathaxe:
We may need some optimization to avoid splitting code too often, when running batch calls to the new methods.

Where would you place global functions?

pygls/workspace.py
```diff
@@ -33,6 +33,58 @@
...
+    line = len(lines)
+    character = 0
+    if line > row:
+        line = row
+        character = sum(1 + int(ord(ch) > 0xFFFF) for ch in lines[line][:col])
+
+    return Position(line, character)
+
+
 class Document(object):
```
rwols, line 88: What's also problematic is that application of these changes is not atomic. You may need to put access to or mutation of the content of the document under a `threading.Lock`, as we seem to be using a threadpool executor.

(By the way, why not use the default threadpool executor from `asyncio`?)

deathaxe: You are refering to the document updates, right?

Yes, reading from stdin seems to be run in a dedicated thread as there seems to be no asyncio endpoint on Windows.

danixeee: Json RPC request messages are collected using threads (as we don't want to reject/miss them), but all _builtin_ methods are executed synnchronously.

So, `textDocument/didChange` is handled in synchronous _buildin_ [method](https://github.com/openlawlibrary/pygls/blob/master/pygls/protocol.py#L609) and unless you are trying to change document content in the _pygls_ workspace from some other method, you are safe.

> (By the way, why not use the default threadpool executor from asyncio?)

Can't remember, but adding _asyncio_ was pretty challenging, specially because it is platform dependent. I was basically learning python through this project and I am aware that code could be improved. We are glad that code quality is being increased by all your contributions.

danixeee:
> Imagine the `_valideate_json()` in your example server was a time expensive asynchronous task which is executed in a separate thread. A language server could have received another didChange notification befor it finishes which updates the document by adding unicode characters to a line which would also be part of the `_validate_json()` command's results.

I see, that makes sense, but I think that core operations (validation, code completions, etc.) should be fast, otherwise your LS would be useless. For custom commands I think developer should decide what to do and if _pygls_ has some barriers, we will find the solution. Thinking about all possible problems in advance is hard and maybe not even needed, but we should stay flexible as much as we can. For e.g., in [this](https://github.com/textX/textX-LS) project even _LanguageServerProtocol_ class is overrided which gives more control if you need it.

> Where would you place global functions?

Maybe you can place them in `workspace.py`?

pygls/workspace.py
deathaxe: Any update of the workspaces content creates a new string instance. Therefore any subsequent function/method using the old instance would keep working with it safely. When done the old reference is deleted and the garbage collector should remove it from memory soon.

Furthermore the `lines` past to the functions are a return value of `splitlines()` so they are independed from the original source anyway.

So even if those functions are not atomic they should not need any `threading.Lock`s from my current understandings.

danixeee:
@deathaxe Can you just update the changelog and rebase/merge with master?

pygls/workspace.py
```diff
@@ -33,6 +33,128 @@
...
+    Arguments:
+        chars (str): The string to return the length in utf-16 code units for.
+    """
+    return len(chars) + utf16_unit_offset(chars)
+
+
+def position_from_utf16(lines: List[str], position: Position) -> Position:
+    """Convert the position.character from utf-16 code units to utf-32.
+
+    A python application can't use the character memeber of `Position`
```
danixeee, line 57: memeber -> member

pygls/workspace.py
```diff
@@ -33,6 +33,128 @@
...
+    directly as per specification it is represented as a zero-based line and
+    character offset based based on a UTF-16 string representation.
```
danixeee, line 59: offset based based on... -> offset based on...

pygls/workspace.py
```diff
@@ -33,6 +33,128 @@
...
+
+    All characters whose codepoint exeeds the Basic Multilingual Plane are
```
danixeee, line 61: exeeds -> exceeds

pygls/workspace.py
```diff
@@ -33,6 +33,128 @@
...
+
+    The offset of the closing quotation mark in x="😋" is
+    - 5 in UTF-16 representation
+    - 4 in UTF-32 representation
+
+    see: https://github.com/microsoft/language-server-protocol/issues/376
+
+    Arguments:
+        lines (list):
+            The content of the document which the position referes to.
```
danixeee, line 72: referes -> refers

pygls/workspace.py
```diff
@@ -33,6 +33,128 @@
...
+            position.character - utf16_unit_offset(lines[position.line][:position.character])
+        )
+    except IndexError:
+        return Position(len(lines), 0)
+
+
+def position_to_utf16(lines: List[str], position: Position) -> Position:
+    """Convert the position.character from utf-32 to utf-16 code units.
+
+    A python application can't use the character memeber of `Position`
```
danixeee, line 91: memeber -> member

pygls/workspace.py
```diff
@@ -33,6 +33,128 @@
...
+    directly as per specification it is represented as a zero-based line and
+    character offset based based on a UTF-16 string representation.
+
+    All characters whose codepoint exeeds the Basic Multilingual Plane are
```
danixeee, line 95: exeeds -> exceeds

pygls/workspace.py
```diff
@@ -33,6 +33,128 @@
...
+
+    The offset of the closing quotation mark in x="😋" is
+    - 5 in UTF-16 representation
+    - 4 in UTF-32 representation
+
+    see: https://github.com/microsoft/language-server-protocol/issues/376
+
+    Arguments:
+        lines (list):
+            The content of the document which the position referes to.
```
danixeee, line 106: referes -> refers

pygls/workspace.py
```diff
@@ -33,6 +33,128 @@
...
+    except IndexError:
+        return Position(len(lines), 0)
+
+
+def range_from_utf16(lines: List[str], range: Range) -> Range:
+    """Convert range.[start|end].character from utf-16 code units to utf-32.
+
+    Arguments:
+        lines (list):
+            The content of the document which the range referes to.
```
danixeee, line 127: referes -> refers

pygls/workspace.py
```diff
@@ -33,6 +33,128 @@
...
+        position_from_utf16(lines, range.end)
+    )
+
+
+def range_to_utf16(lines: List[str], range: Range) -> Range:
+    """Convert range.[start|end].character from utf-32 to utf-16 code units.
+
+    Arguments:
+        lines (list):
+            The content of the document which the range referes to.
```
danixeee, line 145: referes -> refers

danixeee:
@deathaxe Everything works great, can you just address few typos in docstrings and I will approve/merge.

Issue #85, opened 2019-09-30
Implement web-socket server


danixeee:
Currently _pygls_ supports two types of connections:

- TCP/IP
- IO (stdin/stdout)

We want to add another type of connection - _web-sockets_.

bollwyvl:
Sounds cool!

The websocket approach is not precisely standardized: as i mentioned, the convention for how many URLs should be exposed appears to be _n_, one per language server. Having _1_, and doing the negotiation at the message level, sounds really good to me, but there might be deep parts of the protocol model that I don't yet understand!

MatejKastak:
I think the base implementation is resolved in https://github.com/openlawlibrary/pygls/pull/129 . Maybe we can close this issue?

danixeee:
@MatejKastak you are right. Closing the issue, thanks!

PR #129, opened 2020-07-05
Add WebSocket server

MatejKastak:
## Description

Implementation of [WebSocket](https://tools.ietf.org/html/rfc6455) server. This PR allows users to start the WebSocket server on a given `host` and `port` that will "wrap" the LSP protocol. The interface is the same as other `SERVER.start_*` methods.

The main use case is to create convenient interface to interact with editors that are embedded in internet browsers (for example [Monaco LSP-client](https://github.com/TypeFox/monaco-languageclient)), this could potentially allow for broader adoption.

Related to #85 

### DIFF
--- a/pygls/server.py
+++ b/pygls/server.py
@@ -114,7 +114,7 @@ class WebSocketTransportAdapter:
 
     def write(self, data: Any) -> None:
         """Create a task to write specified data into a WebSocket."""
-        asyncio.create_task(self._ws.send(data))
+        asyncio.ensure_future(self._ws.send(data))
 
 
 class Server:
--- a/pygls/server.py
+++ b/pygls/server.py
@@ -216,7 +216,7 @@ class Server:
 
     def start_tcp(self, host, port):
         """Starts TCP server."""
-        logger.info('Starting TCP server on {}:{}'.format(host, port))
+        logger.info('Starting TCP server on %s:%s', host, port)
 
         self._stop_event = Event()
         self._server = self.loop.run_until_complete(
--- a/pygls/server.py
+++ b/pygls/server.py
@@ -24,8 +24,6 @@ from multiprocessing.pool import ThreadPool
 from threading import Event
 from typing import Any, Callable, List, Optional, TypeVar
 
-import websockets
-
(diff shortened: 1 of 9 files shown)


MatejKastak:
I think PR is at the point when it's ready for review/discussion.

1) Adding `websockets` as a dependency made some of the tests fail, how would you like to resolve this problem?
    - `websockets` lib should work on >= python 3.6 based on their docs (We can probably make 3.6 pass)
    - what about python 3.5? You probably want to keed support for 3.5
    - we can make WebSockets feature not default but rather opt-in, and let users install it via `pip install pygls[websockets]` for example. After make some import guard to `start_websocket` and throw runtime error if the package is not found. The tests for this variant would run only on >= 3.6.

2) How to handle `path`argument - [code](https://github.com/openlawlibrary/pygls/pull/129/files#diff-8f1489668ccd105360003e37eba957afR229)
    - we can add optional or madnatory argument to `start_websocket` for user to specify exact path on which the server should listen
    - we can discard it and listen on all endpoints

Any feedback appreachiated.

pygls/server.py
```diff
@@ -197,6 +218,35 @@ def start_tcp(self, host, port):
...
 
+    def start_websocket(self, host, port, path='/'):
+        """Starts WebSocket server."""
+        logger.info('Starting WebSocket server on {}:{}'.format(host, port))
+
+        # TODO: Handle the path
+        self._stop_event = Event()
+        self.lsp._send_only_data = True  # Don't send headers within the payload
+
+        async def connection_made(websocket, cur_path):
```
danixeee, line 229: See above comment. I am not sure about the path argument, I need to take a look at websocket lib.

danixeee (commented):
@MatejKastak Sorry for the delayed response. We have removed support for python 3.5, so that's not a problem anymore. Take a look at the inline comments and let me know what do you think about how we should solve them.

danixeee:
@MatejKastak I just pushed minor updates, thanks and sorry for waiting this long.

danixeee (approved):
Works fine with Monaco client.

Issue #124, opened 2020-05-13
Don't send not specified properties


muffinmad:
E.g.

```python
types.Completion('defaultdict')
```

will result in sending all other attributes along with label:

```
server-reply (id:21) Wed May 13 14:25:56 2020:
(:id 21 :jsonrpc "2.0" :result
     (:isIncomplete :json-false :items
		    [(:label "defaultdict" :kind nil :detail nil :documentation nil :deprecated :json-false :preselect :json-false :sortText nil :filterText nil :insertText nil :insertTextFormat nil :textEdit nil :additionalTextEdits nil :commitCharacters nil :command nil :data nil)]))
```

rwols:
Important notions from the LSP spec:

- There can be an optional key with a valid value.
- There can be a mandatory key with an optional value.

rwols:
There's also a `"range": null` in the response to the textDocument/hover request. I'm guessing it's the same underlying issue.

This library needs to distinguish between **optional keys with typed values** and **mandatory keys**.

rwols:
I haven't tested this but this might work as serializer:

```python
def default_serializer(o):
    """JSON serializer for complex objects."""
    if isinstance(o, enum.Enum):
        return o.value
    d = {}
    for k, v in o.__dict__.items():
        if not k.startswith('_') and v is not None:
            d[k] = v
    return d
```

rwols:
You should also account for cases where a `None` response is allowed:

- textDocument/hover is allowed to respond with `None` if there's no hover.
- textDocument/signatureHelp is allowed to respond with `None` if there's no sighelp.
- ...

rwols:
Another example where this breaks an editor is for textDocument/rename. We do a check here to see if we're dealing with new-style `documentChanges` or old-style `changes`:

```python
def parse_workspace_edit(workspace_edit: Dict[str, Any]) -> Dict[str, List[TextEdit]]:
    changes = {}  # type: Dict[str, List[TextEdit]]
    if 'changes' in workspace_edit:
        for uri, file_changes in workspace_edit.get('changes', {}).items():
            changes[uri_to_filename(uri)] = list(parse_text_edit(change) for change in file_changes)
    if 'documentChanges' in workspace_edit:
        for document_change in workspace_edit.get('documentChanges', []):
            if 'kind' in document_change:
                debug('Ignoring unsupported "resourceOperations" edit type')
                continue
            uri = document_change.get('textDocument').get('uri')
[...]

PR #132, opened 2020-08-20
Add library-specific strategy to address javascript's Optional Properties

pappasam:
## Description
Resolves https://github.com/openlawlibrary/pygls/issues/124  by implementing this suggestion: https://github.com/openlawlibrary/pygls/issues/124#issuecomment-677722383

Basically, this creates a Python-equivalent of typescript `undefined` for this library. `None` == `null`, `undefined` == `undefined`, allowing our serialized messages to more-closely match up with the LSP.

I've tested this E2E with https://github.com/pappasam/jedi-language-server and it works in practice for that language server.

A couple unit tests are now broken (the tests relied on the "bug" identified in https://github.com/pappasam/jedi-language-server). Unfortunately I don't have enough time to revise the tests, but if you like this implementation, I'd merge and have someone more familiar with the tests revise them to more-closely resemble the LSP's spec.

### DIFF
--- a/pygls/protocol.py
+++ b/pygls/protocol.py
@@ -40,7 +40,7 @@ from .types import (ApplyWorkspaceEditParams, ApplyWorkspaceEditResponse, Diagno
                     DidCloseTextDocumentParams, DidOpenTextDocumentParams, ExecuteCommandParams,
                     InitializeParams, InitializeResult, LogMessageParams, MessageType,
                     PublishDiagnosticsParams, RegistrationParams, ServerCapabilities,
-                    ShowMessageParams, UnregistrationParams, WorkspaceEdit)
+                    ShowMessageParams, UnregistrationParams, WorkspaceEdit, undefined)
 from .uris import from_fs_path
 from .workspace import Workspace
 
@@ -74,7 +74,11 @@ def default_serializer(o):
     """JSON serializer for complex objects."""
     if isinstance(o, enum.Enum):
         return o.value
-    return o.__dict__
+    d = {}
+    for k, v in o.__dict__.items():
+        if v is not undefined:
+            d[k] = v
+    return d
 
 
 def deserialize_message(data):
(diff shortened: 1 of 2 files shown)


puremourning:
I just merged this locally after seeing the same bug in `cmake-language-server`

It solves it. Thanks.

Any likelihood of this getting attention from the maintainers ?

danixeee:
@pappasam @puremourning Sorry, I missed to let you know about my [PR](https://github.com/openlawlibrary/pygls/pull/139) that adds type checking and better (de)serialization.

I am waiting to get some feedback on it before I merge it. Please check this [comment](https://github.com/openlawlibrary/pygls/pull/141#issuecomment-750875885).
If you need any help please let me know.

This issue should be resolved there and optional properties won't be sent to the client. Do you still think that we should add logic for `undefined`? What if the language client is not written in javascript and does not have `undefined`?

puremourning:
@danixeee the main problem as it is now is that pygls is sending `"key": null` for optional fields. This is invalid and the keys should _not_ be supplied unless `null` is specified in LSP as a valid value for `key`. That's irrespective of the  client language (my client is python, and expects the key to be supplied or not as in `message.get( "key", default_value )`.

If the other PR is going to resolve that anyway, then great.

danixeee:
@puremourning Ok, I will check it again.

We are now using [pydantic](https://github.com/samuelcolvin/pydantic) for (de)serialization and there's an argument `exclude_unset` which removes keys which values are not provided, and keeps keys even if the value is set to `None`.

Would you have time and the will to test it on your language server?

danixeee:
@puremourning I just pushed [changes](https://github.com/openlawlibrary/pygls/pull/139/commits/9a1fbcc797e3f27b110c17a28bebe1cf96f62637) that should address this issue.

pappasam:
@danixeee amazing, I'm thrilled this is going to be resolved on the next release! Any idea when that will be? I can't wait to use it when it stabilizes.

Issue #118, opened 2020-05-03
Stop once the exit notification is received


rwols:
It's not always the case that this framework shuts down its process once the exit notification is received. In fact once the parent process (the editor) exits, it might get into a busy spin loop, eating up an entire core in my CPU.

I'm fixing killing subprocesses editor-side, but you might want to make sure your process exits anyway, for other editors.

rwols:
I'd suggest just calling `exit(0 if shutdown_received else 1)`.

danixeee:
When I start the server over TCP, I notice that client sends only `shutdown` request and not `exit` notification after it. That happens when I close VS Code window, but I am pretty sure that it was working before. I guess pygls needs to periodically check if client's process is still alive and exit otherwise (#130).

Issue #137, opened 2020-09-29
v0.9.1.


danixeee:
- [x] update versions (@danixeee)
- [x] update CHANGELOG (@danixeee)
- [x] tag/build release (@danixeee )
- [x] upload to PyPi (@danixeee )

Issue #59, opened 2019-03-22
Server capabilities should depend on client capabilities.


danixeee:
The goal is to refactor current implementation of how server capabilities are calculated.
Client capabilities (InitializeRequest) should be considered when creating server capabilities.

https://github.com/openlawlibrary/pygls/blob/master/pygls/types.py#L879

tombh:
Is this still relevant now that v1 is released? Because all types live in `lsprotocol`, do it fix the issue or if not, could it?

alcarney:
It's potentially relevant, while `lsprotocol` provides all the types now it's still on `pygls` to correctly [fill out](https://github.com/openlawlibrary/pygls/blob/master/pygls/capabilities.py#L68) an instance of `ServerCapabilities` as this tells the client which methods etc a server supports. 

While the `ServerCapabilitiesBuilder` does look at some client capabilities, I don't know if it does in all cases etc.

alcarney:
Oh! and we also need to make sure we support any additional fields that were added to the spec since this was last looked at

tombh:
Ok great, I'll look into fixing this soon then.

Issue #130, opened 2020-07-23
This framework doesn't make sure that its process stops once a client crashes or abruptly stops in some way


rwols:
If an editor crashes, it can be the case that the subprocess lingers (becomes a zombie). This framework should ensure that it kills itself, and also any subprocesses it manages.

This is why the `processId` of the parent process (i.e. the editor) is sent along in the initialize request.

danixeee:
Closed via #149.

PR #139, opened 2020-11-19
Added type checking and update types (LSP-3.15)

### LINKED ISSUE #81: LSP method-parameter mappings

danixeee:

- Remove python 3.5 support
- New directory structure for types
- Added new types according to [LSP-3.15 specs](https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/)
- Added type checking (`pydantic`, `typegueard`) for:
  - Feature options
  - Feature params
  - Feature return type
- Store client/server capabilities
- Added a lot of tests

- [ ] test manually with projects that are using `pygls` to make sure that everything is still working

Related to #59, #137 

Closes #81.


renatav (commented):
I know this still a draft pr, but I had the availability today to go through it. Some minor comments. Good job on adding all of those tests!

pygls/capabilities.py
```diff
@@ -0,0 +1,229 @@
...
+                               TEXT_DOCUMENT_DID_SAVE, TEXT_DOCUMENT_WILL_SAVE,
+                               TEXT_DOCUMENT_WILL_SAVE_WAIT_UNTIL, TYPE_DEFINITION,
+                               WORKSPACE_SYMBOL)
+from pygls.lsp.types import (CodeLensOptions, CompletionOptions, DocumentLinkOptions,
+                             ExecuteCommandOptions, ImplementationOptions, SaveOptions,
+                             ServerCapabilities, SignatureHelpOptions,
+                             TextDocumentSyncOptionsServerCapabilities, TypeDefinitionOptions,
+                             WorkspaceFoldersServerCapabilities, WorkspaceServerCapabilities)
+
+# TODO: Check client capabilities also
```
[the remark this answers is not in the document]
danixeee: I'll remove it. I guess that client capabilities should be checked when implementing user defined features.

pygls/capabilities.py
```diff
@@ -0,0 +1,229 @@
...
+
+
```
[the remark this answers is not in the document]
danixeee: I'll add a docstring to `ServerCapabilitiesBuilder` class

pygls/lsp/__init__.py
```diff
@@ -0,0 +1,248 @@
...
+# where:
+#   - CompletionOptions is used when registering a method:
+#   - CompletionParams are received from the client
+#   - Union[List[CompletionItem], CompletionList, None] should be returned by the server (result field of ResponseMessage)
+#
+#       @json_server.feature(COMPLETION, CompletionOptions(trigger_characters=[',']))
+#       def completions(params: CompletionParams = None) -> Union[List[CompletionItem], CompletionList, None]:
+#           """Returns completion items."""
+
+# TODO: support partial-results types
```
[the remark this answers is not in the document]
danixeee: Yes, I haven't had a change to check how it works actually, but I think it's not a big deal, so I will work on it some other time (same is for reporting a progress).

pygls/lsp/types/basic_structures.py
```diff
@@ -73,6 +73,15 @@ class JsonRPCResponseMessage(JsonRpcMessage):
     result: Any
     error: Any
 
+    @root_validator
+    def check_result_or_error(cls, values):
+        result_val, error_val = values.get('result'), values.get('error')
+
+        if result_val is not None and error_val is not None:
+            raise ValueError('Fields "result" and "error" are both given!')
```
renatav, line 81: Both provided or both set sounds better to me

pygls/protocol.py
```diff
@@ -78,12 +79,33 @@ def decorator(self, *args, **kwargs):
...
+    return o.__dict__
+
+
+def deserialize_command(params):
+    """Function used to deserialize command arguments to a specific class
+    or a namedtuple."""
+    cmd_name = params['command']
+    # TODO: Register/Look up custom command arguments' types
+
+    type_name = re.sub('[^a-zA-Z]+', '', cmd_name) + 'Params'
```
renatav, line 102: I understand the general idea, but sure that I understand what needs to be done here. Can you add a comment and provide an example of input and expected output?

danixeee: Updated

tests/lsp/test_completion.py
```diff
@@ -0,0 +1,81 @@
...
+                text_document=TextDocumentIdentifier(uri='file://test.test'),
+                position=Position(line=0, character=0)
+            )
+        ).result(timeout=CALL_TIMEOUT)
+
+        assert response['isIncomplete'] == False
+        assert response['items'][0]['label'] == 'test1'
+        assert response['items'][0]['kind'] == CompletionItemKind.Method
+        assert response['items'][0]['preselect'] == True
+        assert not hasattr(response['items'][0], 'deprecated')
```
krassowski, line 76: Great work on this PR, as an end-user I really look forward to the changes!

Would `'deprecated' not in response['items'][0]` be better here (and in other places where these things are being checked)? I mean `hasattr({'a': 1}, 'a') is False` so this check may be silently passing if the underlying `CompletionItem` is a dictionary (which I would expect).

danixeee: Thanks! Good catch, will fix it.

Have you actually tried out these changes? Did you notice any issues with how it works?

krassowski: No I have not tested it, only reviewed the code to see how it fixes the `null`s. I could try to run it next weekend if you would like some more feedback.

danixeee: I have pushed the fix.

> I could try to run it next weekend if you would like some more feedback.

That would be great. I hope I will find some time to update the docs in the meantime.

danixeee: Did you find time to try it out?

dimbleby:
Is there any prospect of this being merged?  I had a quick play with it, using https://github.com/pappasam/jedi-language-server, seems to work for me.

dimbleby:
Perhaps I should also offer some notes on the experience of switching to this version of pygls in an existing codebase

- Mostly pretty straightforward and mechanical eg changing things from camelCase to snake_case

- jedi-language-server has a couple of particularly ugly bits of code for working round defects in pygls 0.9.1, I was able to remove them entirely, so that's great

- extra typing caught some mistakes that I made in doing the work, also good!

- `ClientCapabilities.has_capability()` is a little awkward: it could do with a type annotation, and it would be nice if it also took a default value to return when the relevant capability was not declared

- there's no help for `InitializeParams.initialization_options`, which I guess makes sense from a spec point of view - it could be literally anything.  But I bet in practice everyone makes it a nested dictionary, so maybe it would be helpful to provide a function analogous to `ClientCapabilities.has_capability()` for extracting values from nested dictionaries?

- mypy is not a fan of all the implicit exports.  In fact generally wouldn't it be nice to be clean under `mypy --strict -p pygls`?  that's probably quite a lot of work, but it's the sort of thing that can be picked away at eg one file at a time...

danixeee:
@dimbleby Great, thanks for the feedback. I will do my best to merge it no later than Monday.

dimbleby:
Super, am excited to see this released!

I have opened #148 against this branch, with some very small suggested tweaks.

docs/source/pages/advanced_usage.rst
```diff
@@ -377,7 +377,7 @@ The code snippet below shows how to send show message log notification:
     async def count_down_10_seconds_non_blocking(ls, *args):
         for i in range(10):
             # Sends message log notification to the client's output channel
-            ls.show_message_log("Counting down... {}".format(10 - i))
+            ls.show_message_log("Counting down... {10 - i}")
```
perrinjerome, line 380:
```suggestion
            ls.show_message_log(f"Counting down... {10 - i}")
```

pygls/protocol.py
```diff
@@ -400,19 +411,25 @@ def _send_response(self, msg_id, result=None, error=None):
...
+        self._send_data(response)
 
     def connection_made(self, transport: asyncio.BaseTransport):
         """Method from base class, called when connection is established"""
         self.transport = transport
 
     def data_received(self, data: bytes):
         """Method from base class, called when server receives the data"""
-        logger.debug('Received {!r}'.format(data))
+        logger.debug('Received {%r}', data)
```
perrinjerome, line 432:
```suggestion
        logger.debug('Received %r', data)
```

pygls/lsp/types/language_features/signature_help.py
```diff
@@ -0,0 +1,86 @@
...
+
+class SignatureHelpInformationClientCapabilities(Model):
+    documentation_format: Optional[List[MarkupKind]] = None
+    parameter_information: Optional[SignatureHelpInformationParameterInformationClientCapabilities] = None
+
+
+class SignatureHelpClientCapabilities(Model):
+    dynamic_registration: Optional[bool] = False
+    signature_information: Optional[SignatureHelpInformationClientCapabilities] = None
+    contextSupport: Optional[bool] = False
```
perrinjerome, line 47: I guess this should be 

```suggestion
    context_support: Optional[bool] = False
```

in other places this is snake case, like here with CompletionClientCapabilities : https://github.com/openlawlibrary/pygls/blob/a5ac2747b741b30e8c04c57bea9cbdcb09d10e73/pygls/lsp/types/language_features/completion.py#L115

pygls/lsp/types/workspace.py
```diff
@@ -0,0 +1,200 @@
...
+
+class DidChangeTextDocumentParams(Model):
+    text_document: VersionedTextDocumentIdentifier
+    content_changes: Union[List[TextDocumentContentChangeEvent],
+                           List[TextDocumentContentChangeTextEvent]]
+
+
+class WillSaveTextDocumentParams(Model):
+    text_document: TextDocumentIdentifier
+    reason: int
```
perrinjerome, line 174: Is this `reason` `int` ? according to https://microsoft.github.io/language-server-protocol/specification#textDocument_willSave it should be `TextDocumentSaveReason`

perrinjerome (commented):
Hello @danixeee , this is great !

I left a few inline comments for minor things.

I also tried to update https://github.com/perrinjerome/vscode-zc-buildout to use this branch. Generally speaking, everything works 👍 here's what I did: first I updated imports, `pygls.types` is now `pygls.lsp.types`, `pygls.features` is now `pygls.lsp.methods`. Then I updated all places where lsp objects where created directly to use keywords arguments in constructors.
I had to change several
```python
Position(1, 2)
```
into
```python
Position(line=1, character=2)
```
because the former is no longer valid. I guess it's OK to break compatibility like this, but in my opinion something should be mentioned in changelog.
 
I also had to change a lot of
```python
Range(Position(1, 2), Position(3, 4))
````
into
```python
Range(start=Position(line=1, character=2), end=Position(line=3, character=4))
````

and at some point I made a mistake, I did something like:

```python
Position(start=1, stop=2)
```

I was disappointed, because this project have a good type coverage, but that error was not reported by mypy. It was also no error at runtime. I looked a bit at pydantic, the models can be configured with `extra = "forbid"` to disallow this. Unfortunately, if we add `extra = "forbid"` in the base model here:

https://github.com/openlawlibrary/pygls/blob/a5ac2747b741b30e8c04c57bea9cbdcb09d10e73/pygls/lsp/types/basic_structures.py#L46-L49

this cause errors during initialization. I did not investigate the error, the message was:
```
pydantic.error_wrappers.ValidationError: 2 validation errors for InitializeParams
capabilities -> textDocument -> completion -> completionItem -> commitCharactersSupport
  extra fields not permitted (type=value_error.extra)
capabilities -> textDocument -> completion -> completionItem -> preselectSupport
  extra fields not permitted (type=value_error.extra)
```
raised here:
https://github.com/openlawlibrary/pygls/blob/a5ac2747b741b30e8c04c57bea9cbdcb09d10e73/pygls/protocol.py#L144

... so it seems we can not use `extra='forbid'` in config. Maybe I missed something here, but I guess if lsp client sends extra properties it should not be an error and these properties should be ignored.

Anyway, good news is that there is a mypy plugin that comes with pydantic, after adding the following mypy config:
```ini
[mypy]
plugins = pydantic.mypy

[pydantic-mypy]
init_forbid_extra = True
```

such errors are reported at type checking time:
```
error: Unexpected keyword argument "start" for "Position"
error: Unexpected keyword argument "end" for "Position"
```

PS: Maybe things would have been better if Position did not have default values for line and characters ? 
https://github.com/openlawlibrary/pygls/blob/a5ac2747b741b30e8c04c57bea9cbdcb09d10e73/pygls/lsp/types/basic_structures.py#L87-L88
there's no such default values in https://microsoft.github.io/language-server-protocol/specification#position

danixeee:
Great catches! I think I have addressed everything you found.

> ... so it seems we can not use extra='forbid' in config. Maybe I missed something here, but I guess if lsp client sends extra properties it should not be an error and these properties should be ignored.

There were additional fields that were wrong and causing validation to fail. Do you think I should keep `extra = "forbid"` in the base model class?

> Maybe things would have been better if Position did not have default values for line and characters?

Not sure if this would help you to detect an issue easier, I will keep default values as they are saying that values are *zero-based*.


@perrinjerome, @dimbleby Could you check everything again and left comments if everything is ok now?

dimbleby:
Latest changes haven't broken anything for me!

I'd be inclined to agree with @perrinjerome that `Position` shouldn't have defaults: I don't know that defaulting it to (0, 0) will help anyone remember that it's zero-based, whereas it definitely provides an opportunity to go wrong.  But I don't feel very strongly about it.

danixeee:
@dimbleby Ok, I think you two are right, we should just follow exact specs...

Do you have any thought about adding `extra = "forbid"` to the base model?

dimbleby:
Again I agreewith @perrinjerome: "forbid" feels dangerous for back-compatibility reasons: if and when they extend the spec so that clients can provide additional fields, it would be a shame for servers to stop working completely for those clients.

(probably such things should be negotiated by capabilities first, but I bet not everyone does that).

danixeee:
Then I think this is ready to be merged. @renatav If you want, you can do a final review, or just approve it.

Regarding the release, I think it's important to investigate and fix #118 and #130 before publishing a new version.

renatav (approved):
The newest changes look good to me. Merge if no further work should be done based on the other feedback

dimbleby:
> Regarding the release, I think it's important to investigate and fix #118 and #130 before publishing a new version.

IMO it would be a shame to block releasing this on those two.  They both are longstanding and as I understand them are quite independent of this rework.  You can always go 0.10.1 after 0.10.0.

Of course if you think that it'll be easy to deal with those two without delaying release enough to care about: then great, more fixes is good!

PR #148, opened 2021-03-13
Lsp miscellany

dimbleby:

Related to #139 (and this pull request is against that branch).

I have:

- allowed pydantic and typeguard versions to go higher.  typeguard declares that it follows semantic versioning, so 2.x should be safe.  pydantic seems to make breaking changes at minor releases; however 1.8 works just fine so I've allowed 1.7.x and 1.8.x

- tweaked `ClientCapabilities.has_capability()` along the lines I suggested in https://github.com/openlawlibrary/pygls/pull/139#issuecomment-798347921 ie added a type annotation, and allowed it to return a default value.  I've also renamed it to `get_capability()`, since it handles not only boolean values (eg `text_document.completion.completion_item.documentation_format` is a list of `MarkupKind`).

### DIFF
--- a/pygls/capabilities.py
+++ b/pygls/capabilities.py
@@ -63,12 +63,12 @@ class ServerCapabilitiesBuilder:
             or TEXT_DOCUMENT_DID_CLOSE in self.features
         )
         will_save = (
-            self.client_capabilities.has_capability(
+            self.client_capabilities.get_capability(
                 'text_document.synchronization.will_save')
             and TEXT_DOCUMENT_WILL_SAVE in self.features
         )
         will_save_wait_until = (
-            self.client_capabilities.has_capability(
+            self.client_capabilities.get_capability(
                 'text_document.synchronization.will_save_wait_until')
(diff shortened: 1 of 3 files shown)


perrinjerome:
Changing the version constraints to be less strict like this makes a lot of sense to me. I feel that libraries should generally not have version constraints in their setup (or maybe a `>= minimal.version` when they require a minimal version) and that picking the good version is a problem for applications. When application depends on multiple libraries, if libraries specify exact requirements in their setup.py it can lead to version conflict. For example, if one wants to build an langage server application with a library "example" and pygls, which has `pydantic==1.7` in their `install_requires` , if that "example" library also depends on pydantic and have a install_requires like `pydantic==1.8`, the installation would be impossible. That's why I believe libraries should not have strict versions in their setup.py. For test requires or docs requires it's not really a problem, but for install requires it can be.

danixeee (approved):
Agreed with everything you said. Thanks for the help! I will merge it now and apply some suggestions from @perrinjerome.

PR #157, opened 2021-03-18
Moved the metadata into setup.cfg

KOLANICH:
## Moved the metadata into setup.cfg

setup.cfg is a declarative format of metadata available since Dec 2016.

So, can be parsed without code execution and can be modified ahtomatically by scripts.

`pyproject.toml` allows the package to be build using `python3 -m build`, also stores settings for `setuptools_scm` that can get and populate precise version information from git.


setup.cfg
```diff
@@ -1,5 +1,72 @@
...
+# Unless required by applicable law or agreed to in writing, software      #
+# distributed under the License is distributed on an "AS IS" BASIS,        #
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. #
+# See the License for the specific language governing permissions and      #
+# limitations under the License.                                           #
+############################################################################
+
+[metadata]
+name = pygls
+version = 0.9.1
```
danixeee, line 20: It's `0.10.1` now.

setup.cfg
```diff
@@ -1,5 +1,72 @@
...
+    Operating System :: OS Independent
+    Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+
+[options]
+packages = find:
+zip_safe = False
+install_requires = pydantic==1.7; typeguard==2.10.0
```
danixeee, line 44: From `setup.py` -> `install_requires=["pydantic>=1.7,<1.9", "typeguard>=2.10.0,<3", "psutil>=5"],`.

danixeee: You can also write each dependency on the new line, for better readability.

setup.py
```diff
@@ -1,102 +1,5 @@
...
-    long_description=long_description,
-    long_description_content_type='text/markdown',
-    url=URL,
-    author=AUTHOR,
-    author_email=AUTHOR_EMAIL,
-    keywords=KEYWORDS,
-    license=LICENSE,
-    packages=packages,
-    include_package_data=True,
-    data_files=[
```
danixeee, line 73: This section is missing in `setup.cfg`, could you add that as well?

danixeee (changes requested):
Please take a look at the inline comments, once you fix them build show work, thanks.

setup.cfg
KOLANICH: Removed `version` at all. It should be auto-detected.

setup.cfg
KOLANICH: Fixed.

setup.py
KOLANICH: `LICENSE.txt` is included automatically, `README.md` goes to `long_description`, `CHANGELOG.md` is appended to `long_description`. Added `ThirdPartyNotices.txt`.

setup.cfg
```diff
@@ -1,5 +1,82 @@
...
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+
+[options]
+packages = find:
+zip_safe = False
+install_requires =
+    pydantic>=1.7,<1.9
+    typeguard>=2.10.0,<3
+    psutil>=5
```
danixeee, line 46: We removed this in the meantime.

danixeee (changes requested):
We had to remove the `psutil` dependency and it's no longer needed, sorry.

`mypy .` command is failing. Could you fix it? It should skip the `./docs` directory, and you would need to fix `server/tests/unit/test_features.py`.

Just replace this:

```python
class FakeServer():
    """We don't need real server to unit test features."""
    publish_diagnostics = Mock()
    show_message = Mock()
    show_message_log = Mock()

    def __init__(self):
        self.workspace = Workspace('', None)
```

and remove lines:

```python
server.publish_diagnostics = Mock()
server.show_message = Mock()
server.show_message_log = Mock()
```

How the version is auto-detected? How can I do major/minor/patch release in case we omit the version from `setup.cfg`?

brettcannon:
I don't see a corresponding issue for discussion, so I guess here is the best place: do you want to stick with setuptools or would you rather to move something simpler like flit? Since the `setup.py` is so simple and there's no C code in pygls you don't have to stick with setuptools if you don't want to.

KOLANICH:
Why to move to `flit_core.buildapi` and `poetry.core.masonry.api` if they both are currently inferior to `setuptools` and don't support `--editable` installs by `pip`? Also eventually [PEP 621](https://www.python.org/dev/peps/pep-0621/) would make all the backends interchangeable from a wheel building point of view.

brettcannon:
> Why to move to flit_core.buildapi and poetry.core.masonry.api if they both are currently inferior to setuptools

I personally don't agree with that sentiment, else I wouldn't have made the suggestion. 😉 

> don't support --editable installs by pip?

Because it's actually not necessary with flit. You can run `flit install` to get the dev dependencies installed and there are various options to get the equivalent of an editable install using a `.pth` file. But since pygls doesn't use a `src/` directory you actually don't need an editable install as `sys.path` will pick up the `pygls/` directory when you run from the top of your checkout; all you need is the dev dependencies installed.

>  Also eventually PEP 621 would make all the backends interchangeable from a wheel building point of view.

Yep, that's why I wrote the PEP. 😉 And flit is working towards supporting PEP 621 already while I don't know when setuptools will get around to it.

I'm not explicitly saying staying with setuptools is a bad idea, I'm just asking if the idea of using a different build tool was considered. But if sticking with setuptools is what the maintainers want then this PR is definitely a good thing as moving away from `setup.py` as much as possible is always a good thing.

KOLANICH:
>But since pygls doesn't use a src/ directory you actually don't need an editable install as sys.path will pick up the pygls/ directory when you run from the top of your checkout; all you need is the dev dependencies installed.

In fact my workflow is to create a dir `forks` somewhere on my machine, to which clone different packages I need, and then install using `-e`. Then just do `git pull`s (and other manipulations via git) mostly without reinstallations.

>Because it's actually not necessary with flit. You can run flit install to get the dev dependencies installed

I prefer to manage dependencies manually. On my machine auto-installation of dependencies (and even resolution of them, and IMHO the new resolver was a big mistake, that can make `pip` as terribly slow as `conda`) is disabled. Instead if I see import errors, I get the packages manually (usually in source form), audit them, if I see something that I deeply don't like (like unnecessary use of `setup.py`, or usage of `pickle`, or other suspicious code, or too lot of large dependencies declared and/or imported) then I fix that for me and create an issue/PR, and install them. This way I get known the packages I have to use better.

IMHO using dedicated tools (and especially the combines like flit and poetry, that depend on lot of packages) is a big step back. The tools should be backends used through unified frontends, and as lightweight as possihle and as modular as possible.

>and there are various options to get the equivalent of an editable install using a .pth file.

Yeah, there exist such options.

danixeee:
@KOLANICH Could you just fix tox?

From the [docs](https://tox.readthedocs.io/en/latest/example/package.html):

```ini
# tox.ini
[tox]
isolated_build = True
```

Hopefully everything else is fine.

KOLANICH:
Recreated `setup.py`. So `isolated_build = True` is not needed.

danixeee (approved):
Great, thanks.

PR #169, opened 2021-04-08
Don't install the tests to site-packages

mattst88:
Commit 8b08673 ("Don't install the tests to site-packages") added
'tests' to the exclude list, but later an 'lsp' subdirectory was added
to tests/ (d0157a5 ("Add test for code completions")). Exclude 'tests*'
to avoid installing 'tests/lsp' to site-packages.


danixeee (approved):
Thanks.

PR #198, opened 2021-06-19
remove defaults from optional fields

dimbleby:
## pygls is serializing too many fields

pygls 0.11.0 serializes certain optional fields, regardless of whether the server has explicitly set them - or whether the client has declared support for them.

Updating [jedi-language-server to take pygls 0.11.0](https://github.com/pappasam/jedi-language-server/pull/151) I see that we are now returning various fields that we previously didn't eg on `CompletionItem` we always set `deprecated` and `preselect`.  I'm pretty sure that this is wrong: not least, there are `deprecatedSupport` and `preselectSupport` on `CompletionClientCapabilities`: if the client has not declared support for these things then presumably we're not supposed to return them.

The approach I have taken is to remove all defaults for Optional fields:
- when the default was not `None`, this had the effect of making the field not optional after all, as above
- when the default was `None` it was redundant to declare it as such, so it's just more compact not to bother

This pull request partially undoes f248002a2fb8a10fb9a5d2bd5a0160a04a6189aa.  I'm not sure what the motivation for that was, but flagging it in case I've broken something.

### DIFF
--- a/pygls/lsp/types/language_features/completion.py
+++ b/pygls/lsp/types/language_features/completion.py
@@ -156,7 +156,7 @@ class CompletionItem(Model):
 
 class CompletionList(Model):
     is_incomplete: bool
-    items: List[CompletionItem] = []
+    items: List[CompletionItem]
 
     def add_item(self, completion_item):
         self.items.append(completion_item)
--- a/pygls/lsp/types/language_features/completion.py
+++ b/pygls/lsp/types/language_features/completion.py
@@ -42,7 +42,7 @@ class CompletionTriggerKind(enum.IntEnum):
 
 class CompletionContext(Model):
     trigger_kind: CompletionTriggerKind
-    trigger_character: Optional[str] = None
+    trigger_character: Optional[str]
 
 
 class InsertTextFormat(enum.IntEnum):
@@ -83,21 +83,21 @@ class CompletionItemKind(enum.IntEnum):
 
 
 class CompletionOptions(WorkDoneProgressOptions):
-    trigger_characters: Optional[List[str]] = None
-    all_commit_characters: Optional[List[str]] = None
-    resolve_provider: Optional[bool] = False
+    trigger_characters: Optional[List[str]]
+    all_commit_characters: Optional[List[str]]
+    resolve_provider: Optional[bool]
 
 
 class CompletionParams(TextDocumentPositionParams, WorkDoneProgressParams, PartialResultParams):
-    context: Optional['CompletionContext'] = None
+    context: Optional['CompletionContext']
 
 
 class CompletionItemKindClientCapabilities(Model):
-    value_set: Optional[List[CompletionItemKind]] = None
+    value_set: Optional[List[CompletionItemKind]]
 
 
 class CompletionTagSupportClientCapabilities(Model):
-    value_set: Optional[List[CompletionItemTag]] = None
+    value_set: Optional[List[CompletionItemTag]]
 
 
 class InsertTextMode(enum.IntEnum):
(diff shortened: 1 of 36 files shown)


dimbleby:
I found a handful more places where pygls inserts defaults that are not given by the LSP specification, and removed them.

If this second commit is more controversial than the first then happy to separate the two, but hopefully that's not the case.

pygls/lsp/types/language_features/completion.py
```diff
@@ -136,27 +136,27 @@ class InsertReplaceEdit(Model):
...
+    additional_text_edits: Optional[List[TextEdit]]
+    commit_characters: Optional[List[str]]
+    command: Optional[Command]
+    data: Optional[Any]
 
 
 class CompletionList(Model):
     is_incomplete: bool
-    items: List[CompletionItem] = []
+    items: List[CompletionItem]
```
danixeee, line 159: Could you please check and initialize `items` filed in `add_item` and `add_items` methods?

pygls/lsp/types/language_features/document_highlight.py
```diff
@@ -57,4 +57,4 @@ class DocumentHighlightKind(enum.IntEnum):
 
 class DocumentHighlight(Model):
     range: Range
-    kind: Optional[DocumentHighlightKind] = DocumentHighlightKind.Text
+    kind: Optional[DocumentHighlightKind]
```
danixeee, line 60:
```typescript
export interface DocumentHighlight {
	/**
	 * The range this highlight applies to.
	 */
	range: Range;

	/**
	 * The highlight kind, default is DocumentHighlightKind.Text.
	 */
	kind?: DocumentHighlightKind;
}
```

It says "The highlight kind, default is DocumentHighlightKind.Text", does it mean that if the server does not send this field, the client will use this value, or the server should send `DocumentHighlightKind.Text` ?

danixeee (commented):
Please check inline comments, I am not sure about default values from the specs.

pygls/lsp/types/language_features/completion.py
dimbleby: I don't _think_ this should be necessary.  How has the user constructed a `CompletionList` on which to call these methods without initializing the `items`?

Can you show me some code that goes wrong without the change that you suggest?

pygls/lsp/types/language_features/document_highlight.py
dimbleby: If it meant the second of those, then the field would not be optional - right?

So I definitely read this the first way.

pygls/lsp/types/language_features/completion.py
danixeee: I guess you are right, I am still not used to pydantic.

pygls/lsp/types/language_features/document_highlight.py
danixeee: Probably, ok merging this.

danixeee (approved):
Works good, thanks!

perrinjerome:
> This pull request partially undoes https://github.com/openlawlibrary/pygls/commit/f248002a2fb8a10fb9a5d2bd5a0160a04a6189aa. I'm not sure what the motivation for that was, but flagging it in case I've broken something.

@dimbleby  it seems this broke progress notification https://github.com/openlawlibrary/pygls/issues/231 . It seems to be that in this pull request all default values were removed, so maybe it would also be OK to "keep" this code in `__init__` ?

dimbleby:
I think it would be preferable to use pydantic features.  Currently we explicitly exclude defaults from serialization - https://github.com/openlawlibrary/pygls/blob/720da69128654fa9c98f0e917451eef76cc9a0d8/pygls/protocol.py#L382.

I expect that not doing that should address this?

That might result in a few unwanted `None` in the serialization, but either:
- track down the places where they are explicitly set to None, and don't do that
- or lazy approach, `exclude_none=True`

PR #229, opened 2022-03-29
updates pydantic to resolve #221

### LINKED ISSUE #221: Why Pydantic?
Is there any reason why you use Pydantic internally? Compared to attrs+cattrs it's at least 5 times slower and doesn't provide any benefits.

https://stefan.sofa-rockers.org/2020/05/29/attrs-dataclasses-pydantic/
https://threeofwands.com/why-i-use-attrs-instead-of-pydantic/

kammoh:
update pydantic (and other dependencies, including test, doc, dev, etc)

Fixes  #221

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,17 +20,24 @@ classifiers =
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
+    Programming Language :: Python :: 3.11
 
 [options]
-setup_requires = setuptools>=44; wheel; setuptools_scm>=3.4.3; toml
+setup_requires =
+    setuptools>=45
+    wheel
+    setuptools_scm[toml]>=6.2
 packages = find:
 zip_safe = False
-install_requires = pydantic>=1.7,<1.9; typeguard>=2.10.0,<3
+install_requires =
+    pydantic~=1.9
+    typeguard~=2.13
 include_package_data = True
 tests_require =
-    mock==3.0.5
-    pytest==4.5.0
-    pytest-asyncio==0.10.0
+    mock==4.0.3
+    pytest==7.0.1
+    pytest-asyncio==0.18.3
 
 [options.packages.find]
 exclude =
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -60,3 +69,6 @@ max-line-length = 99
 
 [mypy]
 ignore_missing_imports = True
+
+[tool:pytest]
+asyncio_mode = auto
```
perrinjerome, line 74: according to my understanding of https://github.com/openlawlibrary/pygls/blob/4772250f8ede25aee8e457dae5bca394aeccdd2f/.editorconfig#L7 there should be a new line at end of file here

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{36,37,38,39}
+envlist = py{36,37,38,39,310,311}
```
perrinjerome, line 2: I believe we also need to update the list of python versions here so that they are tested on github actions
https://github.com/openlawlibrary/pygls/blob/4772250f8ede25aee8e457dae5bca394aeccdd2f/.github/workflows/ci.yml#L16

perrinjerome (commented):
This generally looks good ( this is also what I felt in https://github.com/openlawlibrary/pygls/pull/148#issuecomment-799020887 )

I left a few remarks, I don't know if it's OK for me to click approve, I'm just an external contributor

tox.ini
kammoh: Thank you @perrinjerome for the feedback. I tried to fixed the issues. Tox tests run fine with local Pyenv versions, and at least [esbonio LS](https://github.com/swyddfa/esbonio) (based on pygls) seems to be functioning correctly with the local installation of the PR.

setup.cfg
```diff
@@ -39,16 +45,18 @@ exclude =
 
 [options.extras_require]
 ws =
-    websockets==9.*
+    websockets~=10.2
```
alcarney, line 48: I think websockets `10.x` drops support for Python 3.6, this might have to stay on `9.x` for now

arwedus: maybe it's okay to drop 3.6 support as well for pygls? Would be cool to see this PR get forward.

alcarney: Perhaps, though I think that call probably lies with @danixeee  or @dgreisen

Bobronium: Is it necessary to update `websockets` to support python 3.10? 


If not, I'd propose doing it in a separate PR, while merging this first, to address compatibility issues. 

Then we can reason about whether dropping python3.6 is a good idea.

Issue #243, opened 2022-06-29
pydantic throws ValidationError


slackline:
I've encountered an error that pydantic has with `pygls` when used by `jedi-language-server`...

```Traceback (most recent call last):
  File "/home/neil/.local/lib/python3.9/site-packages/pygls/protocol.py", line 152, in deserialize_message
    deserialize_params(data, get_params_type)
  File "/home/neil/.local/lib/python3.9/site-packages/pygls/protocol.py", line 138, in deserialize_params
    data['params'] = params_type(**params)
  File "pydantic/main.py", line 406, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for InitializeParams
processId
  value is not a valid integer (type=type_error.integer)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/neil/.local/bin/jedi-language-server", line 8, in <module>
    sys.exit(cli())
  File "/home/neil/.local/lib/python3.9/site-packages/jedi_language_server/cli.py", line 125, in cli
    SERVER.start_io()
  File "/home/neil/.local/lib/python3.9/site-packages/pygls/server.py", line 204, in start_io
    self.loop.run_until_complete(
  File "/usr/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/home/neil/.local/lib/python3.9/site-packages/pygls/server.py", line 75, in aio_readline
    proxy(b''.join(message))
  File "/home/neil/.local/lib/python3.9/site-packages/pygls/protocol.py", line 456, in data_received
    json.loads(body.decode(self.CHARSET),
  File "/usr/lib/python3.9/json/__init__.py", line 359, in loads
    return cls(**kw).decode(s)
  File "/usr/lib/python3.9/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.9/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
  File "/home/neil/.local/lib/python3.9/site-packages/pygls/protocol.py", line 154, in deserialize_message
    raise JsonRpcInvalidParams()
pygls.exceptions.JsonRpcInvalidParams: Invalid Params
```

[...]

Issue #234, opened 2022-05-29
Crashes on Python 3.11


Bobronium:
Official support for python 3.10 was added only to pydantic in version `1.9`, however in lib requirements lower version is pinned:
https://github.com/openlawlibrary/pygls/blob/4ce8988c87ba75352a510611881bc1d9d4e6757f/setup.cfg#L28

Might be fixed in #229
Causes https://github.com/microsoft/vscode-black-formatter/issues/67

Initially I thought it crashes on Python 3.10, but as it turned out, I got it wrong.

karthiknadig:
@tombh @alcarney Can this issue be prioritized, this is currently impacting about 8 different servers? We might need a temporary solution until pygls 1.0.0 can hit final release.

tombh:
I have time today, I'll look into it and either fix it or find a temporary workaround.

BTW, what are those servers?

karthiknadig:
Following are from Microsoft: 
- https://github.com/microsoft/vscode-isort
- https://github.com/microsoft/vscode-black-formatter
- https://github.com/microsoft/vscode-pylint
- https://github.com/microsoft/vscode-autopep8
- https://github.com/microsoft/vscode-flake8
- https://github.com/microsoft/vscode-python-tools-extension-template (uses pygls as foundation to create extensions for vscode for tools in python)
There are others that are from community: jedi-language-server, vscode-ufmt

tombh:
Quick question, it seems like the fix is to set `pydantic>=1.10.2` in order to include their https://github.com/pydantic/pydantic/pull/4012 PR. It seems easy, so I'm wondering if this wasn't suggested before because upgrading Pydantic would break other project's dependencies?

BTW thanks for the list, could we add them to https://github.com/openlawlibrary/pygls/blob/master/Implementations.md ?

karthiknadig:
Looking at other issues, it seems like potential breaking changes in `pydantic` were the reason.

tombh:
So how about if Pygls set its dependency to `pydantic>=1.19.1,<=1.10.2`? I'm not familiar enough with Python to know if that's helpful? The idea being that most users will remain unaffected, but that if somebody really needs to use Python 3.11 then they'll need to explicitly install Pydantic `1.10.2`.

brettcannon:
> So how about if Pygls set its dependency to `pydantic>=1.19.1,<=1.10.2`? I'm not familiar enough with Python to know if that's helpful?

[...]

Issue #101, opened 2020-01-28
Create new package for protocol-level concerns


bollwyvl:
> Moved here from https://github.com/openlawlibrary/pygls/issues/93#issuecomment-573763184

In a nutshell:

- complete the [derivation of a JSON Schema](https://gist.github.com/bollwyvl/7a128978b8ae89ab02bbd5b84d07a4b7) for the Language Server Protocol 
- Use the schema to generate a new, standalone package, e.g. `language-server-protocol` 
  - with good typing, e.g. dataclasses, typeddict or namedtuple
  - either
    - offer multiple versions of the library, e.g. `language-server-protocol==3.14.x` (only one per env)
    - include multiple versions of the spec e.g. currently `lsp.v3_14` (would allow multiple servers with different versions)
  - offer optionally pluggable serialization (e.g. `json`, `orjson`, `ujson`) and runtime validation (e.g. `jsonschema`, `fastjsonschema`)
- use `language-server-protocol` in `pygls`
  - remove most of the constants and classes of `pygls.types` and `pygls.protocol`

For a language server implementer, the default would look the same:

```py
# setup.py
    install_requires = ["pygls"],

# server.py
import pygls
server = pygls.LanguageServer()

@server.feature(server.lsp.SELECTION_RANGE)
def foo():
    return None  # would fail because SELECTION_RANGE not defined
```

> no validation, stdlib `json`, use the latest stable LSP

While a fancier implementation might choose:

```py
# setup.py
    install_requires = ["pygls", "orjson", "fast-jsonschema"],

# server.py
import language_server_protocol
import pygls
lsp = language_server_protocol.V3_15(
    json=lsp.io.OrJSON(), 
    validator=lsp.schema.FastJSONSchema()
)

server: LanguageServer[lsp.V3_15] = LanguageServer(lsp=lsp)

@server.feature(lsp.SELECTION_RANGE)
async def bar():
    return await some_untyped_function()  # would fail at runtime if `range` not defined in result
```

> use bleeding edge protocol, hotrod rust-based JSON, and do runtime checking with a code-generating parser

bollwyvl:
Discussed off-line: before starting we should resolve some plumbing:

> -  creating another organisation/repository
> - creating a separate project inside the same repository with pygls (as Microsoft did with their nodejs LSP implementation)?

I'm going to assume the latter means: "in-tree", e.g. another `src`. Having a long-running orphan branch is the fastest to start, but has most of the disadvantages of both.

[...]

PR #273, opened 2022-10-17
v1.0.0 BREAKING CHANGES release thread

tombh:
With huge thanks and kudos to [Esbonio](https://github.com/swyddfa/esbonio)'s @alcarney and Microsoft's @karthiknadig we now have a pre release version of Pygls that migrates away from Pydantic and to [lsprotocol](https://github.com/microsoft/lsprotocol).

⚠️ This involves breaking changes. Existing projects depending on Pygls will need to update their code. Testers are welcome.

You can try it by explicitly setting your dependency to `pygls=1.0.0a2`. The migration guide is at: [docs/pages/migrating-to-v1.rst](https://github.com/openlawlibrary/pygls/blob/v1-lsprotocol-breaking-alpha/docs/source/pages/migrating-to-v1.rst)

`lsprotocol` is mostly just a module that officially and comprehensively defines the data structures and types that make up the LSP specification. Therefore, what it offers are upstream (from Microsoft, the creators of LSP) guarantees on adherence to the formal behaviour and expectations of the specification. 

TODO:
* [x] ~~Write~~ Publish guide on how to make the required changes.
* [x] Isolate a commit in a sample LSP server project that demonstrates the changes. Eg; [jedi-language-server already has a PR](https://github.com/pappasam/jedi-language-server/pull/230/commits). Or maybe make a verbose commit in the new ["Pygls Starter project"](https://github.com/tombh/cli-tools-lsp) I'm working on.
* [x] Get feedback from people that have successfully made the migration:
  * https://github.com/pappasam/jedi-language-server/pull/230.
  * https://github.com/charliermarsh/vscode-ruff/pull/37
* [x] @alcarney what else would you like to see happen before the final release?
* [x] Remove advice to start new projects with v1 alpha

This may also be a good opportunity to introduce other breaking changes, such as in #274

Notes:
* Issues that this release should either address or fix:
  * #221
  * #243 
  * #234 
  * #101
  * #275
* See [the original "lsprotocol" PR](https://github.com/openlawlibrary/pygls/pull/264) for further discussion.

Edit: added line about taking advantage of a breaking change release


dimbleby:
My main observation from https://github.com/pappasam/jedi-language-server/pull/230 is that it's a real shame that lsprotocol doesn't expose types, that will make authoring language servers that little bit more error prone than it need be.

Otherwise it all seems to have gone pretty smoothly, nice job.

karthiknadig:
@dimbleby if you see anything missing, please file bug on `lsprotocol`. I can try and improve code generation to produce any missing or supporting types.

dimbleby:
@karthiknadig you probably want to get involved in https://github.com/pappasam/jedi-language-server/pull/230?notification_referrer_id=NT_kwDOAA1asLE0NjMzNDU0MTUzOjg3NTE4NA#issuecomment-1282891441

tombh:
@dimbleby Wow that's awesome that you've already managed to integrate this. Can you explain more about lsprotocol not exposing types? I mean I see that it does use Python types, eg: [`types.py`](https://github.com/microsoft/lsprotocol/blob/main/lsprotocol/types.py). So do you mean that it lacks a `py.typed` stub to help tools like Mypy orient themselves? Or do you mean that it doesn't give type-based feedback during serialisation/deserialisation? Or something else?

dimbleby:
I mean type annotations, I would love for lsprotocol to (i) be clean under `mypy --strict` and (ii) include a `py.typed` so that users could take advantage of that

you'll likely have seen that it's not quite so awesome as all that, https://github.com/pappasam/jedi-language-server/pull/230?issuecomment-1282891441 looks like an lsprotocol bug

tombh:
Ok, so it's more a matter of having _better_ typing support, rather than introducing types in the first place? It seems like `lsprotocol` very much leans in the direction of wanting to fully support Mypy right? So it should be a legitimate next step to explore, or even that @karthiknadig intends for full support and we've just surfaced some minor bugs?

dimbleby:
some way still to go, if mypy support is intended:
```
$ mypy lsprotocol
...
Found 25 errors in 2 files (checked 5 source files)

$ mypy --strict lsprotocol
...
Found 68 errors in 4 files (checked 5 source files)
```
Actually that's a relatively small number for a project that perhaps hasn't yet even tried - possibly it _wouldn't_ be tremendously hard to get to zero errors.

karthiknadig:
Created https://github.com/microsoft/lsprotocol/issues/96 to track `mypy --strict` issues.

alcarney:
Not sure how we want to handle issues with this branch? Do we want to create dedicated issues but tag them with something like `v1`, or just track everything in this PR?

I've noticed that the server process just completely crashes if it encounters a message it cannot parse/does not like.
Happy to open a dedicated issue for this if we want to go that route :)

alcarney:
Not sure how we want to handle issues with this branch? Do we want to create dedicated issues but tag them with something like `v1`, or just track everything in this PR?

I've noticed that the server process just completely crashes if it encounters a message it cannot parse/does not like.
Happy to open a dedicated issue for this if we want to go that route :)

tombh:
Good question @alcarney, it probably depends on the issue. I think that one may need its own issue, because it's more likely to be a bug in Pygls itself right?

tombh:
Just a little update to summarise from the other threads involved.

* @karthiknadig released a new version of `lsprotocol` that fixes an issue that was identified in the Jedi server integration https://github.com/pappasam/jedi-language-server/pull/230#issuecomment-1282891441
* @dimbleby [experimented with adding a `py.typed` marker file to their local copy of `lsprotocol`](https://github.com/pappasam/jedi-language-server/pull/230#issuecomment-1287875155) and successfully got typing information. This is good news for https://github.com/microsoft/lsprotocol/issues/96

alcarney:
@tombh yes, the error handling in `pygls` needs some refinement as one bad message probably shouldn't take out the server completely

tombh:
I've made a new issue for the more general problem of unhandled errors: https://github.com/openlawlibrary/pygls/issues/277

tombh:
I've just rebased the recent changes we've merged into master in the last few days, most notably improved error handling #282. That addresses an underlying issue that @alcarney noticed with this new code, namely that the server wasn't catching unhandled deserialisation errors, which I think understandably could come up more now that we've changed all the underlying types.

But I think @alcarney, you still need to fix that error? All we've done with #282 is catch unhandled errors, rather than fix the unhandled error right?

Also, where are we at with the Mypy types? I don't think that's necessarily a blocking issue? The main thing is that https://pypi.org/project/lsprotocol/2022.0.0a6 fixed the errors in https://github.com/pappasam/jedi-language-server/pull/230#issuecomment-1282891441 right?

alcarney:
> But I think @alcarney, we still need to fix that error?

The error in question was being tracked in microsoft/lsprotocol#102, and should be fixed in the next version of `lsprotocol`

tombh:
Ah great! I'll look out for that then

alcarney:
Hmm... I wonder why all the builds have hung... 

> I think the only main thing left to do is write a brief v0.12.x to v1 migration guide.

I can put something together as I have a go at porting esbonio - unless anyone else has started working on one?

> So it looks like we're getting there with this branch

Exciting! Though do we know how many servers have attempted a migration?

tombh:
No, nobody else (well it would only be me anyway) has started on a migration guide. I think you'd be best placed to write it, so if you're up for that then great, but I'm totally happy to do it too.

The only report from the wild about implementing this migration so far that we've had is https://github.com/pappasam/jedi-language-server/pull/230 Yeah, would be nice to get some more feedback for sure.

So, the CI errors. There are 3 interesting things happening.

1. The LSP client that we use in tests, which is repurposed from the LSP server, doesn't (really?? still!?) have a central place to timeout asyncio futures 🫤 Hence the hung CI. Point 3 here in particular is the culprit for the hang. Let me think about a way to refactor the LSP test client so that at the very least CI doesn't hang again like this.

    PS. I assume that the fact that it's the unique circumstances of the repurposed LSP test client, that this hanging behaviour doesn't reflect a potential issue with the actual server itself. I'm not 100% sure though. I just assume that the server doesn't call anything like a timeout-less, blocking `Future.result()` in non-async functions 👀

2. `CodeActionKind.Refactor` isn't parsing:
```
E       AssertionError: assert 'refactor' == <CodeActionKind.Refactor: 'refactor'>
E        +  where 'refactor' = CodeAction(title='action2', kind='refactor', diagnostics=None, is_preferred=None, disabled=None, edit=None, command=None, data=None).kind
E        +  and   <CodeActionKind.Refactor: 'refactor'> = CodeActionKind.Refactor
```

3. `FoldingRange` isn't parsing:
Can be isolated with: `.tox/py/bin/python3.10 -m pytest -vvv tests/lsp/test_folding_range.py::test_folding_range_return_list --log-cli-level=DEBUG`
```
| Traceback (most recent call last):
|   File "<cattrs generated structure lsprotocol.types.FoldingRange>", line 28, in structure_FoldingRange
|     res['kind'] = __c_structure_kind(o['kind'], __c_type_kind)
|   File "/home/streamer/Workspace/pygls/.tox/py/lib/python3.10/site-packages/cattrs/converters.py", line 377, in _structure_error
|     raise StructureHandlerNotFoundError(msg, type_=cl)
| cattrs.errors.StructureHandlerNotFoundError: Unsupported type: typing.Union[lsprotocol.types.FoldingRangeKind, str, NoneType]. Register a structure hook for it.
| Structuring class FoldingRange @ attribute kind
```

alcarney:
> 1. The LSP client that we use in tests, which is repurposed from the LSP server, doesn't (really?? still!?) have a central place to timeout asyncio futures 

Fun fact, the test client in `pytest-lsp` was originally based on the LSP client in pygls' test suite, it's just been adapted to talk to a subprocess rather than an `os.pipe()` and I'm facing the exact same issue. I've been playing around with a few ideas (the most promising having a background thread cancel all pending tasks when the subprocess exits), but nothing that works in all cases... 

Anyway if you're interested [here](https://github.com/alcarney/lsp-devtools/tree/pygls-1.0) is my current WIP branch for `pytest-lsp`  maybe we'll be able to figure something out together on this.

*Side note: Do you think `pytest-lsp` is something that should "live" as part of pygls? Maybe eventually be used to do end to end tests on pygls itself??*

> 2. CodeActionKind.Refactor isn't parsing:

Interesting... I assume it's related to the fix for microsoft/lsprotocol#101

alcarney:
> No, nobody else (well it would only be me anyway) has started on a migration guide. I think you'd be best placed to write it, so if you're up for that then great, but I'm totally happy to do it too.

I'm writing a migration guide as I go, hopefully `esbonio` touches enough of pygls' API that it results in a decent first draft :) 

--------

On a separate note, an issue I'm having porting `esbonio` is with the `WORKSPACE_DID_DELETE_FILES` notification, `pygls` is rejecting my feature registration

```python
    @server.feature(
        WORKSPACE_DID_DELETE_FILES,
        FileOperationRegistrationOptions(
            filters=[
                FileOperationFilter(
                    pattern=FileOperationPattern(glob="**/*.rst"),
                )
            ]
        ),
    )
    def on_delete_files(ls: RstLanguageServer, params: DeleteFilesParams):
```
with the error 
```python
  File "/usr/lib64/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib64/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/var/home/alex/Projects/esbonio-pygls/lib/esbonio/esbonio/__main__.py", line 18, in <module>
    sys.exit(main())
  File "/var/home/alex/Projects/esbonio-pygls/lib/esbonio/esbonio/__main__.py", line 14, in main
    cli.main(_cli)
  File "/var/home/alex/Projects/esbonio-pygls/lib/esbonio/esbonio/cli.py", line 90, in main
    server = create_language_server(args.server_cls, modules)
  File "/var/home/alex/Projects/esbonio-pygls/lib/esbonio/esbonio/lsp/__init__.py", line 128, in create_language_server
    return _configure_lsp_methods(server)
  File "/var/home/alex/Projects/esbonio-pygls/lib/esbonio/esbonio/lsp/__init__.py", line 182, in _configure_lsp_methods
    def on_delete_files(ls: RstLanguageServer, params: DeleteFilesParams):
  File "/var/home/alex/Projects/pygls/pygls/feature_manager.py", line 174, in decorator
    raise TypeError(
TypeError: Options of method "workspace/didDeleteFiles" should be instance of type <class 'lsprotocol.types.FileOperationOptions'>
```
`pygls` gets the type to check against from the `METHOD_TO_TYPES` dict in `lsprotocol`, which as far as I can tell aligns with what I'm passing to the `@feature()` decorator
```python
# In lsprotocol/types.py
    WORKSPACE_DID_DELETE_FILES: (
        WorkspaceDidDeleteFilesNotification,
        None,
        DeleteFilesParams,
        FileOperationRegistrationOptions,
    ),
```
So I'm not entirely sure where `pygls` is getting `lsprotocol.types.FileOperationOptions` from.

I am working around #278 at the moment so I suspect this may also be relevant... 
```python
class Patched(LanguageServerProtocol):
    """Tweaked version of the protocol allowing us to tweak how the `ServerCapabilities`
    are constructed."""

    def __init__(self, *args, **kwargs):
        self._server_capabilities = ServerCapabilities()
        super().__init__(*args, **kwargs)

    @property
    def server_capabilities(self):
        return self._server_capabilities

    @server_capabilities.setter
    def server_capabilities(self, value: ServerCapabilities):

        if WORKSPACE_DID_DELETE_FILES in self.fm.features:
            opts = self.fm.feature_options.get(WORKSPACE_DID_DELETE_FILES, None)
            if opts:
                value.workspace.file_operations.did_delete = opts  # type: ignore

        self._server_capabilities = value
```

Hopefully, I'm just missing something obvious somewhere and I'll spot my mistake next time I look at it :smile:

karthiknadig:
With this one:`AssertionError: assert 'refactor' == <CodeActionKind.Refactor: 'refactor'>` it is basically comparing a string with an enum. Either we can do  `CodeActionKind.Refactor.value` or I can update the CodeActionKind to derive from (str, Enum), which should resolve that.

I am investigating the folding range one.

karthiknadig:
Found the issues, will make a fix soon in `lsprotocol`:
image

tombh:
I made [an issue on lsp-devtools](https://github.com/alcarney/lsp-devtools/issues/24) to talk about exploring a solution to the hangs together.

@alcarney I've looked at your `WORKSPACE_DID_DELETE_FILES` issue, but I'm afraid I'm not familiar enough with it to have any immediately useful feedback.

@karthiknadig great thanks, I'll look out for the release.

dimbleby:
Using a typechecked lsprotocol claims some bugs in this repository:
```
$ mypy pygls
pygls/workspace.py:194: error: Item "TextDocumentContentChangeEvent_Type2" of "Union[TextDocumentContentChangeEvent_Type1, TextDocumentContentChangeEvent_Type2]" has no attribute "range"
pygls/workspace.py:257: error: Item "TextDocumentContentChangeEvent_Type2" of "Union[TextDocumentContentChangeEvent_Type1, TextDocumentContentChangeEvent_Type2]" has no attribute "range"
pygls/protocol.py:794: error: "Type[TraceValues]" has no attribute "VERBOSE"
pygls/progress.py:78: error: Argument "value" to "ProgressParams" has incompatible type "WorkDoneProgressBegin"; expected "Union[LSPObject, List[Any], str, int, int, float, bool, None]"
pygls/progress.py:83: error: Argument "value" to "ProgressParams" has incompatible type "WorkDoneProgressReport"; expected "Union[LSPObject, List[Any], str, int, int, float, bool, None]"
pygls/progress.py:86: error: Argument "value" to "ProgressParams" has incompatible type "WorkDoneProgressEnd"; expected "Union[LSPObject, List[Any], str, int, int, float, bool, None]"
Found 6 errors in 3 files (checked 11 source files)
```

the first three look like real bugs, the second three look like false positives (perhaps the lsprotocol type annotations aren't as helpful as they might be around `LSPAny`).

tombh:
Thanks for those type errors @dimbleby. Indeed, the first 3 should be easy fixes for Pygls.

@karthiknadig do you have any thoughts about those `WorkDoneProgressXXX` errors? I couldn't find anything in `lsprotocol` that inherits from `LSPObject`, so should `WorkDoneProgressXXX` values be passed as lists in order to match the `List[Any]` part of the `Union`?

karthiknadig:
`LSPObject` is supposed to represent an object with undefined properties. This is how the LSP sepec model defines it:
```
        {
            "name": "LSPObject",
            "properties": [],
            "documentation": "LSP object definition.\n@since 3.17.0",
            "since": "3.17.0"
        },
```

In the case of `ProgressParams.value` it is set to `LSPAny` so it could take any object. But looks like the definition is not describing that well. I will investigate.

karthiknadig:
I just published a bunch of fixes. Can you update and re-try?

dimbleby:
@karthiknadig going from 2022.0.0a7 to 2022.0.0a8 introdues unit test failures in this repository:
```
>       assert actual == expected
E       AssertionError: assert {'method': 'test/notification', 'jsonrpc': '2.0', 'params': {'token': 'id1', 'value': {'title': 'Begin progress', 'percentage': 0}}} == {'jsonrpc': '2.0', 'method': 'test/notification', 'params': {'token': 'id1', 'value': {'kind': 'begin', 'percentage': 0, 'title': 'Begin progress'}}}
E         Common items:
E         {'jsonrpc': '2.0', 'method': 'test/notification'}
E         Differing items:
E         {'params': {'token': 'id1', 'value': {'percentage': 0, 'title': 'Begin progress'}}} != {'params': {'token': 'id1', 'value': {'kind': 'begin', 'percentage': 0, 'title': 'Begin progress'}}}
E         Full diff:
E           {
E            'jsonrpc': '2.0',
E            'method': 'test/notification',
E            'params': {'token': 'id1',
E         +             'value': {'percentage': 0,
E         -             'value': {'kind': 'begin',
E         -                       'percentage': 0,
E                                 'title': 'Begin progress'}},
E           }
```

karthiknadig:
Fix provided by @alcarney  https://github.com/microsoft/lsprotocol/pull/122

karthiknadig:
Try with `2022.0.0a9`. Once this branch is merged, I will add `pygls` tests to `lsprotocol` CI and we can prevent any further regressions.

tombh:
Just released https://pypi.org/project/pygls/1.0.0a3 it should fix all typing errors and #285 and #286.

alcarney:
As of #288, I was able to successfully get a version of esbonio running on the new version of pygls (swyddfa/esbonio#484) so including #290, I've made all the technical changes I can think of.

There's still some cleanup needed in the docs however, (checking code examples, removing references to pydantic etc)

Other than that though, what will our next steps be? Would it be worth seeing if we can get some of the other PRs/issues resolved ready for the big 1.0? (e.g. #253, #230)

tombh:
I think your integration of this into esbonio is one of the best indications we could have that this nearing release quality. I know you probably haven't covered the entire LSP spec, but considering that you also implemented this branch I feel you have the best overview.

I think we can only wait for more server authors to update and give feedback. But even without that I think we're in a good a good enough place to justify a formal release. No rush though.

#253 and #230 would certainly be nice to have in a v1 release, but they're not necessarily related to the new lsprotocol dep right?

alcarney:
> but considering that you also implemented this branch I feel you have the best overview.

Also the least likely to spot a mistake 😉 

> but they're not necessarily related to the new lsprotocol dep right?

Correct, just a nice to have as you said

charliermarsh:
If it's at all helpful to see, here's the first-pass on upgrading [Ruff's](https://github.com/charliermarsh/ruff) VS Code extension to `1.0.0a3`: https://github.com/charliermarsh/vscode-ruff/pull/37.

tombh:
@charliermarsh Yes, that is indeed helpful to see! Thank you ❤️

Everybody: I've updated Pygls' README and docs to state that v1 alpha is now the recommended version to start new Pygls-based LSP servers.

Having released a fix on our main branch for a LSP type issue that is already fixed in this branch, I'm feeling like we're ready to release this branch.

Today, the fact that this branch is not released caused a tiny bit more real extra work rather than the theoretical extra work that might come from releasing this branch. Which is to say, there's a reasonable assumption that this branch will not cause any extra work when released, and an actual fact that not releasing it has already created a tiny bit more work.

Fixing the LSP type issue today felt very much like supporting an old version of Pygls. Which is something we can still totally do even once this v1 branch is released.

What do you think @alcarney?

alcarney:
I agree, I can't think of a technical reason to hold this branch back anymore.

The only other things that's probably worth doing before a release, is to make sure the docs, README etc are consistent with all the changes here.

tombh:
Great. I've just pushed those doc changes and bumped the version to `v1.0.0`. I'm ready to merge! Sanity-check depending of course.

alcarney:
Looks good.

One final thought, should we wait for a non-alpha release of `lsprotocol`?

brettcannon:
> One final thought, should we wait for a non-alpha release of `lsprotocol`?

FYI Karthik won't be available to make a final release until the latter half of this month, but we would be happy to if the library is working as you all need it to be.

tombh:
@alcarney Good point. Ideally yes, it'd be good to release with a non-alpha `lsprotocol`. I wonder if we're in a bit of a chicken and the egg situation? That's the main reason I'm voting for a release now, just to fall on one side of the line. It's only a vote, so totally happy to wait as well. But I'm feeling like this v1 branch now offers something better than current v0.x. Even if it might not be perfect.

@brettcannon From my perspective I think `pygls` and `lsprotocol` are ready and am happy to even release `pygls` with the alpha `lsprotocol`. So take that as one vote of confidence, but of course let the general consensus be the guide.

alcarney:
@tombh, just remembered another change I think we wanted to make so I've opened #295.

Happy to release `v1.0` with an alpha `lsprotocol` if you are, just thought I'd check :slightly_smiling_face:

tombh:
Well, LGTM then 🚀

alcarney (approved):
Can't believe it's actually happening! :sweat_smile:

tombh:
Woohoo! 🎉 Merged and published. Thank you everybody ❤️🙇

I actually streamed the moment live on Twitch 😎 https://clips.twitch.tv/DeadExquisiteButterPJSugar-DLXXPnknw87VSdPs

Here's the Pypi package: https://pypi.org/project/pygls/1.0.0

And the migration docs have been published too: https://pygls.readthedocs.io/en/latest/pages/migrating-to-v1.html (I notice I didn't get the list syntax right for the links to other already migrated projects, I'll make a little PR for that).

charliermarsh:
Congratulations on the release!

brettcannon:
Congrats on the release! And thanks to everyone for providing feedback on lsprotocol so we could improve it for everyone's benefit!

PR #279, opened 2022-10-24
fix: Require Pydantic 1.10.2 when Python is 3.11

tombh:
This finally gives us Python 3.11 support.

See #234 for more discussion.

~~- [ ] [CONTRIBUTORS.md][contributors] was updated, as appropriate~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev"]
 
     runs-on: ${{ matrix.os }}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning][semver].
 ### Changed
 ### Fixed
 
+## [0.12.3] - 24/10/2022
+### Fixed
+- Require Pydantic 1.10.2 when Python is 3.11
+
 ## [0.12.2] - 26/09/2022
 ### Fixed
 - Relaxed the Python version upper bound to `<4`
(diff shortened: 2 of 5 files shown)


karthiknadig (approved):

Bobronium (approved):

alcarney (approved):

brettcannon:
If I'm reading this correctly, there is no single version of Pydantic that pygls can use which is compatible with all supported versions of Python? It looks like you have it set so that Python 3.10 and older use Pydantic 1.9.1 or 1.9.2, while Python 3.11 is 1.10 or newer. Is Pydantic 1.10 not usable for Python 3.7 - 3.10?

brettcannon:
Or put another way, upper bounds on dependencies is typically considered bad practice for libraries. I can find you a blog post reference if you want all the gory details.

tombh:
I fully agree it's bad practice. The main excuse is that we are close to releasing a major version bump that completely removes Pydantic: #273. It certainly may be possible to release Pygls with a single version of `pydantic==1.10.2`, but [Pydantic v1.10.0 introduced breaking changes](https://pydantic-docs.helpmanual.io/changelog/) (doing so on minor version increments is itself generally argued as bad practice right?). Considering those points and the immanent release of Python 3.11 (which simply did not work at all before this PR), it seems better to just, hopefully temporarily, manage with this working, albeit bad practice, solution.

I'm curious if those points didn't come across clearly in #234? If I misunderstood something or moved too hastily, it'd be good to know. At the very least I would hope that Pygls isn't adding more unnecessary inconveniences of bad practice to the ecosystem.

brettcannon:
> The main excuse is that we are close to releasing a major version bump that completely removes Pydantic: #273.

Right, but you haven't released it yet and it seems far out enough you are making this fix. 😉 If you weren't bothering with the fix I would understand, but if you are then folks like us are going to take this fixed release and bump up against issues if you install a single version of pygls to service multiple versions of Python (e.g. we install a copy copy of pygls for our Black formatter extension in VS Code that needs to work across Python 3.7 - 3.11; this PR technically means we can't get a version of pydantic in there that meets that criteria with pygls).

PR #280, opened 2022-10-24
Remove upper bound on pydantic for python <3.11

karthiknadig:

Currently the restriction on pydantic does not allow us select a single version that works for python 3.7-3.11. This PR removes the upper bound on python version <3.11. This allows pygls to select a single version of pydantic that works on all python versions.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ python_requires = >=3.7,<3.12
 packages = find:
 zip_safe = False
 install_requires =
-    pydantic>=1.9.1,<1.10  ; python_version<"3.11"
+    pydantic>=1.9.1 ; python_version<"3.11"
     pydantic>=1.10.2 ; python_version>="3.11"
     typeguard>=2.10.0,<3
 include_package_data = True
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning][semver].
 ### Changed
 ### Fixed
 
+## [0.12.4] - 24/10/2022
+### Fixed
+- Remove upper bound on Pydantic when Python is <3.11
+
 ## [0.12.3] - 24/10/2022
 ### Fixed
 - Require Pydantic 1.10.2 when Python is 3.11
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -8,6 +8,7 @@
 - [DeathAxe](https://github.com/deathaxe)
 - [Denis Loginov](https://github.com/dinvlad)
 - [Jérome Perrin](https://github.com/perrinjerome)
+- [Karthik Nadig](https://github.com/karthiknadig)
 - [Laurence Warne](https://github.com/LaurenceWarne)
 - [Matej Kašťák](https://github.com/MatejKastak)
 - [Max O'Cull](https://github.com/Maxattax97)
--- a/pygls/__init__.py
+++ b/pygls/__init__.py
@@ -19,7 +19,7 @@
 import os
 import sys
 
-__version__ = "0.12.3"
+__version__ = "0.12.4"
 
 IS_WIN = os.name == 'nt'
 IS_PYODIDE = 'pyodide' in sys.modules


tombh:
Thanks for this, it directly addresses @brettcannon's [comment](https://github.com/openlawlibrary/pygls/pull/279#issuecomment-1289646484) in #279.

All the tests pass and everything looks good.

I'm just curious, as someone not so familiar with Python dependencies. Does this change theoretically make it possible that a server author depending on Pygls could see issues with their server when upgrading to Pygls 0.12.4? Is it possible that a server author who hasn't pinned their own version of Pydantic and has unfortunately implemented Pydantic-based code suffers from the breaking changes in Pydnatic's `1.10.0`. I feel like it's quite unlikely but from my current understanding I can't guarantee that it's impossible.

brettcannon:
> Is it possible that a server author who hasn't pinned their own version of Pydantic and has unfortunately implemented Pydantic-based code suffers from the breaking changes in Pydnatic's `1.10.0`.

If it's just 1.10.0 that's the problematic release then the < Python 3.11 requirement can expilcitly avoid that single version. I'll leave a review comment for Karthik to that effect if I'm understanding correctly.

setup.cfg
```diff
@@ -33,7 +33,7 @@ python_requires = >=3.7,<3.12
 packages = find:
 zip_safe = False
 install_requires =
-    pydantic>=1.9.1,<1.10  ; python_version<"3.11"
+    pydantic>=1.9.1 ; python_version<"3.11"
```
brettcannon, line 36:
```suggestion
    pydantic>=1.9.1,!=1.10.0 ; python_version<"3.11"
```

karthiknadig: the breaking change in 1.10.0 are there in 1.10.1 and 1.10.2. So this restriction does not address that. But it seems the pygls is not using the APIs that actually were broken by 1.10.0 from what I can tell.

tombh: That's right. Pygls isn't using those APIs that broke in Pydantic v1.10.0, and that's proved by our test suite matrix running against Python v3.11rc2. But I'm not clear how we prove that there aren't Pygls custom server developers out there that _are_ depending on those Pydantic APIs? As far as I can see there's no way to prove that, as unlikely as it might be.

brettcannon: > I'm not clear how we prove that there aren't Pygls custom server developers out there that _are_ depending on those Pydantic APIs?

Does it matter? As in if someone used those versions of Pydantic and used those broken APIs would it break pygls itself, or just their own code? If it's the latter then I would argue that's their business and not pygls'; Python is a language for consenting adults and if someone wants to use a version of pydantic with some bugs that's their choice.

tombh:
Bringing out the code comment conversation to here...

I have to say that this perspective is quite new to me and I'm having a bit of trouble with it. I'm certainly not the most experienced Python dev, but neither am I a beginner. Whether rightly or wrongly I've never been advised on the necessity to pin versions of implicitly installed modules that I "borrow" from an explicitly installed dependency. My understanding, which I suspect now might be misinformed, is that sub-dependencies are, in affect, a part of the module. So in our case I do consider Pydantic to be part of Pygls, in a sense it's the "Pygls-sanctioned version of Pydantic".

As such, I see relaxing the upper limit of Pydantic as a breaking change that I personally would associate with a major version bump to Pygls: simply because there is a known, non-zero chance that it will break other projects that depend on Pygls. However, I do also realise that Pygls can't be expected to bump its own major version number whenever any of its own dependencies make breaking changes! Although Pydantic is unique in our case because it's such a fundamental and indeed _public_ part of the Pygls API, as in the fact that Pydantic methods are exposed on Pygls objects via `pydantic.BaseModel`'s public methods.

I don't at all mean to say that I won't support this PR. I just want to understand the reasoning.

Quoting from @brettcannon in #279:
> folks like us are going to take this fixed release [that fixes an upper Pydantic limit] and bump up against issues if you install a single version of pygls to service multiple versions of Python (e.g. we install a copy of pygls for our Black formatter extension in VS Code that needs to work across Python 3.7 - 3.11; this PR technically means we can't get a version of pydantic in there that meets that criteria with pygls).

I'm not totally clear on that example. Let's say that vscode-black would like to support the new Python version. In order to currently do that it has to specify `pygls=0.12.3` which, depending on the user's current system Python version, will install differing versions of Pydantic. The problem comes when there's another Pygls-based LSP server that hasn't upgraded to `0.12.3` yet and the end user has Python 3.11 as their system Python. Because now there is no single version of Pygls/Pydantic that works on all their Pygls-based LSP servers. I don't think that's the example you're giving is it?

alcarney:
> My understanding, which I suspect now might be misinformed, is that sub-dependencies are, in affect, a part of the module

Perhaps the way to look at it is, if a language server depends on Pydantic for more than just Pygls, enough to be affected by whatever breaking change is in 1.10 (I haven't looked into the details yet), then Pydantic really should be promoted to a top level dependency for that project. 

> I don't think that's the example you're giving is it?

From what I understand (though I could be wrong), extensions like vscode-black bundle the Python dependencies for their language servers, so are looking for a consistent set of versions that work across OSes, python versions etc.

brettcannon:
> sub-dependencies are, in affect, a part of the module

Sort of. They definitely get installed, but only so much as what is required to run the module. Passed that, to be blunt, it isn't your concern/business what the user wants to install. Or put another way, your business/opinion stops at what pygls requires to run properly.

> So in our case I do consider Pydantic to be part of Pygls, in a sense it's the "Pygls-sanctioned version of Pydantic".

I wouldn't word it as "sanctioned" as much as "known to be compatible". It is up to pip (or whatever installer you use) to calculate what packages will meet the set of constraints to get all of the packages installed while all being compatible with each other. Because Python uses a flat package hierarchy you can't police other people's dependencies like this without unnecessarily restricting what they can do.

Remember, Python is a language of consenting adults and that extends to the design of its packaging ecosystem (at least in a PyPA sense; can't speak for conda). If someone is fine with pydantic 1.10.0 and it doesn't break pygls, that's their choice to use that version.

> there is a known, non-zero chance that it will break other projects that depend on Pygls

... if they _choose_ to use Pydantic in a way that triggers the bug. As Alex said:

> if a language server depends on Pydantic for more than just Pygls, enough to be affected by whatever breaking change is in 1.10 (I haven't looked into the details yet), then Pydantic really should be promoted to a top level dependency for that project.

It is up to projects to specify their runtime dependencies appropriately, but not to make recommendations of what they think is appropriate beyond that. Once again, to be blunt, it's none of your concern/business if/why I want to use a buggy version of Pydantic for my needs. As long as pygls still works appropriately for that version then it's my choice to use the buggy version.

Trust me, people have weird/odd/unique reasons to use a specific version of a dependency (a simple one is the security team has approved pydantic 1.10.0 but no other version, or legal has only cleared that version of pydantic). Artificially restricting it indirectly because pygls is trying to protect people from themselves will just cause grief for those folks stuck in a situation where they have no choice over what version of pydantic they get to use. If pydantic thought the release was that bad they could yank it from PyPI so only people explicitly asking for that version would get it installed.

> In order to currently do that it has to specify `pygls=0.12.3` which, depending on the user's current system Python version, will install differing versions of Pydantic.

> From what I understand (though I could be wrong), extensions like vscode-black bundle the Python dependencies for their language servers, so are looking for a consistent set of versions that work across OSes, python versions etc.

That's totally correct. We ship a single copy of pygls for all versions of Python that we support thanks to it being in pure Python.


> does anybody have a link to any of those VSCode extensions' Github repos or similar?

An example is https://github.com/microsoft/vscode-black-formatter which was built from https://github.com/microsoft/vscode-python-tools-extension-template acting as a template.

tombh:
Thank you so much for your detailed reply. And apologies that I hadn't understood this area of package management quite as well as one would hope for a maintainer of a relied upon module.

I've publish this PR  as 0.12.4 to PyPI.

I'm now also slightly worried that I hadn't come across those vscode-extension repos before. They're really, really important to Pygls. Firstly it's important to see how a major editor is using and relying upon Pygls. Why weren't they added to [Implementations.md](https://github.com/openlawlibrary/pygls/blob/master/Implementations.md)? But also that python tools template one really should be directly linked to from our README as the de facto way to run Pygls servers in VSCode. I see it's only a 7 month old repo, maybe that's why I hadn't heard of it before?

brettcannon:
> Thank you so much for your detailed reply.

Quite welcome! I have to get something out of all the years I have sunk into Python packaging. 😉



> And apologies that I hadn't understood this area of package management quite as well as one would hope for a maintainer of a relied upon module.

No worries! You are not the first, nor only person to not know every single subtle detail to Python packaging.



> Why weren't they added to [Implementations.md](https://github.com/openlawlibrary/pygls/blob/master/Implementations.md)?

We honestly didn't think about it / knew about that file. Plus the language servers for our tools extensions are embedded in the repo and not a standalone language server like the other projects listed in that file.

If you would like us to list all of our tools extensions using pygls in that file we can send a PR to add them.





> I see it's only a 7 month old repo, maybe that's why I hadn't heard of it before?

It also only got [announced in August](https://devblogs.microsoft.com/python/python-in-visual-studio-code-august-2022-release/#python-tools-extension-template).

> But also that python tools template one really should be directly linked to from our README as the de facto way to run Pygls servers in VSCode.

We can send a PR for that as well if you want.

tombh:
Haha thanks for your encouragement 😊

No worries about not having updated about the extensions here. I was worried that I'd just missed something about it, a issue or PR or message or something. I've made a note about it: https://github.com/openlawlibrary/pygls/issues/281

PR #282, opened 2022-10-25
fix: Improved error handling

### LINKED ISSUE #277: Improve handling of unexpected errors
This jumps off from an issue identified by @alcarney in the v1 breaking changes PR https://github.com/openlawlibrary/pygls/pull/273#issuecomment-1288146841

Pygls does already catch some errors, but not all. Nor does it offer a way for server authors to handle those unexpected errors.

@alcarney Is that a fair summarisation of the problem? Obviously the underlying message parsing problem needs to be fixed, but is it not a reflection of a deeper issue, that Pygls couldn't recover from that error?

Recently I've been attempting to build a Pygls template project, that offers a reasonable and informative starting point for new custom servers. Maybe it's my relative inexperience with Pygls, but I noticed that there doesn't seem to be a centralised point for all error handling? My naive approach is to [decorate the `feature` decorator like so](https://github.com/tombh/cli-tools-lsp/blob/development/src/lsp/server.py#L53):
```python
def add_feature(self, feature: str):
    """
    This is a wrapper just to catch and handle all unexpected errors in one place.
    TODO:
    Is this even useful? I wonder if this should be formally supported in Pygls itself?
    """

    def wrapper(func):
        @self.feature(feature)
        async def inner(*args, **kwargs):
            try:
                await func(*args, **kwargs)
            except Exception as e:
                self.logger.error(e)
                # TODO: should this have the name of the custom LSP server?
                self.show_message(f"Unexpected error in LSP server: {e}")

        return inner

    return wrapper
```

Does that seem relevant to the deeper issue at hand?

tombh:
This includes a notable, perhaps even controversial, behaviour change:
All errors other than LSP requests are now sent to the client as `showMessage.type = MessageType.Error` messages. This may result in existing custom LSP servers showing errors that were not seen before, these aren't new errors, just previously undisplayed errors.

Another advantage to this is that it is now easier to test error handling. When running end to end tests, all errors are now accessible in the client.

Also, `JsonRPCProtocol.data_receieved()` now catches unhandled errors.

Fixes #277.

### DIFF
--- a/docs/source/pages/advanced_usage.rst
+++ b/docs/source/pages/advanced_usage.rst
@@ -447,6 +447,19 @@ And method invocation example:
 
     server.send_notification('myCustomNotification', 'test data')
 
+Custom Error Reporting
+^^^^^^^^^^^^^^^^^^^^^^
+
+By default Pygls notifies the client to display any occurences of uncaught exceptions in the
+server. To override this behaviour define your own `report_server_error()` method like so:
+
+.. code:: python
+
+    Class CustomLanguageServer(LanguageServer):
+        def report_server_error(self, error: Exception, source: Union[PyglsError, JsonRpcException]):
+            pass
+
+
 Workspace
 ~~~~~~~~~
 
--- a/pygls/protocol.py
+++ b/pygls/protocol.py
@@ -21,7 +21,6 @@ import json
 import logging
 import re
 import sys
-import traceback
 import uuid
 from collections import namedtuple
 from concurrent.futures import Future
@@ -33,7 +32,8 @@ from pygls.capabilities import ServerCapabilitiesBuilder
 from pygls.constants import ATTR_FEATURE_TYPE
 from pygls.exceptions import (JsonRpcException, JsonRpcInternalError, JsonRpcInvalidParams,
                               JsonRpcMethodNotFound, JsonRpcRequestCancelled,
-                              MethodTypeNotRegisteredError)
+                              MethodTypeNotRegisteredError, FeatureNotificationError,
+                              FeatureRequestError)
 from pygls.feature_manager import (FeatureManager, assign_help_attrs, get_help_attrs,
                                    is_thread_function)
 from pygls.lsp import (JsonRPCNotification, JsonRPCRequestMessage, JsonRPCResponseMessage,
@@ -316,8 +316,14 @@ class JsonRPCProtocol(asyncio.Protocol):
             self._execute_notification(handler, params)
         except (KeyError, JsonRpcMethodNotFound):
             logger.warning('Ignoring notification for unknown method "%s"', method_name)
-        except Exception:
-            logger.exception('Failed to handle notification "%s": %s', method_name, params)
+        except Exception as error:
+            logger.exception(
+                'Failed to handle notification "%s": %s',
+                method_name,
+                params,
+                exc_info=True
+            )
+            self._server._report_server_error(error, FeatureNotificationError)
 
     def _handle_request(self, msg_id, method_name, params):
         """Handles a request from the client."""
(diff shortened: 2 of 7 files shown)


pygls/protocol.py
```diff
@@ -392,8 +396,14 @@ def _send_data(self, data):
                 self.transport.write(header + body)
             else:
                 self.transport.write(body.decode('utf-8'))
+            print(data)
```
perrinjerome, line 399:
```suggestion
```

docs/source/pages/advanced_usage.rst
```diff
@@ -447,6 +447,19 @@ And method invocation example:
...
+Custom Error Reporting
+^^^^^^^^^^^^^^^^^^^^^^
+
+By default Pygls notifies the client to display any occurences of uncaught exceptions in the
+server. To override this behaviour define your own `report_server_error()` method like so:
+
+.. code:: python
+
+    Class CustomLanguageServer(LanguageServer):
+        def report_server_error(self, message: str):
```
perrinjerome, line 459: how about a signature like this ?
 
```py
        def report_server_error(self, error: Exception):
```

then custom server implementations would have more context.

pygls/protocol.py
```diff
@@ -427,6 +437,17 @@ def connection_made(self, transport: asyncio.BaseTransport):
         self.transport = transport
 
     def data_received(self, data: bytes):
+        try:
+            self._data_received(data)
+        except Exception:
+            logger.error(traceback.format_exc())
```
perrinjerome, line 443: I know that this pattern is present in other places, but instead of:
```py
            logger.error(traceback.format_exc())
```
we can use logging ability to log the currently handled exception with something like this:

```py
            logger.error("Error receiving data", exc_info=True)
```

docs/source/pages/advanced_usage.rst
alcarney: Seconded :)

alcarney:
Even though there is a mechanism for sending an error response for JSON-RPC requests, do you think it's worth calling `report_server_error` anyway? It feels odd to have a function that's called for any error.... except this one scenario.

It might also be worth adding a second argument to `report_server_error` that tells it which situation it's being called in.
I imagine the default implementation would want to handle all cases apart from requests.

pygls/protocol.py
tombh: Updated in latest force push

docs/source/pages/advanced_usage.rst
tombh: Thirded!

pygls/protocol.py
tombh: Deleted, thanks

tombh:
I've updated the commit:
* Replaced `message: str` arg with `error: Exception` arg in `report_server_error()`. But still not sending that error to the editor by default (a plain human-readable message is sent instead). So its use case is more for authors that want to override the default `report_server_error()`.
* `report_server_error()` now takes a `source` arg so that request errors can also be potentially dual-reported, though their sending as `showMessage` errors is suppressed by default (current behaviour of including any errors in the LSP response object remains unchanged). At first I thought an enum would be good. Then I thought custom errors, which is a bit weird because you'd think that you could get that custom error from the `error` arg, but of course these are unexpected errors so the underlying exception type could be anything. So I've created these 2 new error types anyway: `FeatureRequestError(PyglsError)` and `FeatureNotificationError(PyglsError)`. Which is indeed a little curious because they're never raised, they're only used for matching the error source. But it feels better to create them rather than enums, because at least these can have other use cases the code base.
* Also some other refactoring. And a test to check that custom `report_server_error()` methods can inadvertently create recursive error reports.

pygls/server.py
```diff
@@ -426,6 +433,37 @@ def show_message_log(self, message, msg_type=MessageType.Log) -> None:
...
+        and so can seemingly be sent at random. Also for example consider JSON RPC serialization
+        and deserialization, if a payload cannot be parsed then the whole request/response cycle
+        cannot be completed and so one of these "out of band" error messages is sent.
+
+        These "out of band" error messages are not a requirement of the LSP spec. Pygls simply
+        offers this behaviour as a recommended default. It is perfectly reasonble to override this
+        default.
+        """
+
+        if isinstance(source, FeatureRequestError):
```
alcarney, line 462: Since the error classes rather than instances are being passed to `source`, I believe this check should be `issubclass` instead

alcarney:
Made a suggestion on the if check in the default `report_server_error` implementation, but otherwise looking good!

tombh:
Oh! Yes, I should have known that 🤦 And in fact, the test that was testing for that was a false positive, so I wrote an extra test that tests that test hahah

tombh:
Oh, actually, saying that, Mypy wasn't happy and actually `if source == FeatureRequestError:` seems to work. Make sense?

alcarney (dismissed):

tombh:
Awesome, thanks for the feedback everybody. LGTM 🚀. Just added the CHANGELOG deets, can I get another review acceptance please 🙇

alcarney (approved):

PR #291, opened 2022-11-18
Add pygls playground?

alcarney:

image

I thought a fun way we could celebrate the 1.0 release would be to also provide a "playground" where you can go to a webpage and try `pygls` right there in your browser - thanks to our pyodide support!

This is just a proof of concept, to check that the idea can work and to check **if we even want to accept this into the pygls repo?**

This will easily take as much, if not more effort to build and maintain as `pygls` itself.

There are also some limitations to this setup that would prevent you from trying every feature of pygls / the lsp spec, but it would hopefully serve as a good intro at least.

Or should we explore implementing this as a VSCode web extension? At least we wouldn't have to worry about the UI side as much - though it would come with its own set of technical tradeoffs.

Let me know your thoughts :smile: 


tombh:
Goodness me 😲🤯 This is amaaazing. It'd be great to have this in, alongside, or somehow associated with Pygls in whatever way. Ideally it'd live under a pygls-org namespace, with various other projects, even the JSON example I think would be better as its own repo. Anyway, that's another conversation.

So does the Monaco editor have its own LSP client?

alcarney:
> So does the Monaco editor have its own LSP client?

Not a first party one, but there is [this project](https://github.com/TypeFox/monaco-languageclient) that covers a portion of the spec.

tombh:
Oh right, that's pretty well developed. So what would you consider the largest part of the work needed here?

alcarney:
Definitely the UI.

It's already functional, in that you can write a server on the right and have it work with what you do on the left.

However, it won't tell you about any syntax errors/runtime errors in your code etc. 

Then there's plenty you could do to make the experience nicer / more intuitive, e.g log output, LSP traffic visualisation the list goes on :) 

As much as I like the idea of it being "just" a webpage, I do wonder if it would be better in the long run to make it a VSCode web extension - just from the sheer amount of stuff we'd get for free...

tombh:
What would you consider a usable MVP of this? Is it enough just to display the STDOUT/STDERR of the server?

By VSCode _web_ extension, you mean do all this in the version of VSCode that runs in the browser? Would that be able to support playground IDs? Eg; https://pyglsplayground.com/abc123 So individual playgrounds could be shared?

alcarney:
> What would you consider a usable MVP of this?

If it gave you some feedback on your syntax/runtime errors that's probably good enough to start with.
Oh, and maybe something that looks a bit nicer too :smile: 

> you mean do all this in the version of VSCode that runs in the browser?

Yes.

> So individual playgrounds could be shared?

It's one of the things we'd get for free, any repo on GitHub could become a playground and accessible via a URL like
https://vscode.dev/github/openlawlibrary/pygls

That's not to say it would be a seamless experience though:
- Users would have to install the web extension themselves (though it's possible to get the UI to recommend it to you)
- Users would have to login with a GitHub account to access any repos - I assuming so it can determine if you're able to commit back to the repo or not

Hmm... :thinking: after writing those two points out, perhaps it's best to start with the webpage version, especially since it's halfway there already?

tombh:
So the MVP needs 3 areas; client, server and output. Makes sense.

Yeah, I think that too. VSCode in the browser is definitely very powerful, but I think what you've built already is more in the spirit of a playground. It's something that people don't think twice about loading up. It's something quick and easy to play around with a little idea and share it with others.

alcarney:
@tombh I assume this got closed since the base branch got deleted?

tombh:
Oh! I forgot about that. So yes it was closed automatically. I still totally think this would be awesome. Can you change the base branch to master?

alcarney:
I don't see an option in the UI... I think I'll have to recreate it

PR #298, opened 2022-12-04
some type-checking fixes

dimbleby:
A handful of type-checking fixes.

There's lots more that could be done - try `mypy --strict pygls` if you want to see.  But this covers enough of the public API that typechecking from jedi-language-server is happy.

### DIFF
--- a/pygls/protocol.py
+++ b/pygls/protocol.py
@@ -27,7 +27,7 @@ from collections import namedtuple
 from concurrent.futures import Future
 from functools import lru_cache, partial
 from itertools import zip_longest
-from typing import Any, Callable, List, Optional, Type, Union
+from typing import Any, Callable, List, Optional, Type, TypeVar, Union
 
 import attrs
 from cattrs.errors import ClassValidationError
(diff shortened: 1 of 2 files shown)


tombh (dismissed):

tombh:
Thank you. Do you need these in a release? Would you like to add the conventional commit tag, bump the patch version and add a quick note to the CHANGELOG? I can get it released straight away then.

And yes, `mypy --strict` support is definitely something I'd like to see at some point.

dimbleby:
No hurry at all, I'm just passing the time and making myself useful

tombh:
Could you rebase latest master? Then I'll get this merged. I know we don't yet require commit message syntax. But could you add a `chore:` please. [I'd like for us to start requiring Conventional Commit syntax](https://github.com/openlawlibrary/pygls/issues/300#issuecomment-1338172108) in order to auto generate the CHANGELOG file.

tombh (approved):

tombh:
Great, thank you ✨

PR #303, opened 2022-12-14
fix: True LSP Diagnostic param, with `version`

### LINKED ISSUE #211: Add an optional version parameter to LanguageServerProtocol publish_diagnostics
- The LSP v15 added an optional version parameter to the publishDiagnostics notification.
- PublishDiagnosticsParams includes version but,
- LanguageServerProtocol.publish_diagnostics is not  using version

Please make version an optional parameter of LanguageServerProtocol.publish_diagnostics so that Lsp Servers can implement this feature.

tombh:
Fixes #211

Includes deprecation warning for old method signature

I don't know if I'm being too clever here?

Also, does anybody know why there are often 2 public methods to do the same thing like this? Therefore, we can call both `pygls.server.LanguageServer.publish_diagnostics` and `pygls.protocol.LanguageServerProtocol.publish_diagnostics`. I assume it's because `pygls.server.LanguageServer` is the more friendly user-facing class, whilst `pygls.protocol.LanguageServerProtocol` is internal and has to be comprehensive.

### DIFF
--- a/pygls/server.py
+++ b/pygls/server.py
@@ -408,9 +408,23 @@ class LanguageServer(Server):
         """Gets the object to manage client's progress bar."""
         return self.lsp.progress
 
-    def publish_diagnostics(self, doc_uri: str, diagnostics: List[Diagnostic]):
-        """Sends diagnostic notification to the client."""
-        self.lsp.publish_diagnostics(doc_uri, diagnostics)
+    def publish_diagnostics(
+        self,
+        uri: str,
+        diagnostics: Optional[List[Diagnostic]] = None,
+        version: Optional[int] = None,
+        **kwargs
+    ):
+        """
+        Sends diagnostic notification to the client.
+        """
+        params = self.lsp._construct_publish_diagnostic_type(
+            uri,
+            diagnostics,
+            version,
+            **kwargs
+        )
+        self.lsp.publish_diagnostics(params, **kwargs)
 
     def register_capability(self, params: RegistrationParams,
                             callback: Optional[Callable[[], None]] = None) -> Future:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,11 @@ and this project adheres to [Semantic Versioning][semver].
  - Fix progress example in json extension. ([#230])
  - Fix `AttributeErrors` in `get_configuration_async`, `get_configuration_callback`, `get_configuration_threaded` commands in json extension. ([#307])
  - Fix type annotations for `get_configuration_async` and `get_configuration` methods on `LanguageServer` and `LanguageServerProtocol` objects ([#307])
+ - Provide `version` param for publishing diagnostics ([#303]) 
 
 [#230]: https://github.com/openlawlibrary/pygls/issues/230
+[#303]: https://github.com/openlawlibrary/pygls/issues/303
 [#307]: https://github.com/openlawlibrary/pygls/issues/307
-
 
 ## [1.0.0] - 2/12/2022
 ### Changed
--- a/pygls/protocol.py
+++ b/pygls/protocol.py
@@ -835,10 +835,71 @@ class LanguageServerProtocol(JsonRPCProtocol, metaclass=LSPMeta):
 
         self.notify(LOG_TRACE, params)
 
-    def publish_diagnostics(self, doc_uri: str, diagnostics: List[Diagnostic]) -> None:
-        """Sends diagnostic notification to the client."""
-        self.notify(TEXT_DOCUMENT_PUBLISH_DIAGNOSTICS,
-                    PublishDiagnosticsParams(uri=doc_uri, diagnostics=diagnostics))
+    def _publish_diagnostics_deprecator(
+        self,
+        params_or_uri: Union[str, PublishDiagnosticsParams],
+        diagnostics: Optional[List[Diagnostic]],
+        version: Optional[int],
+        **kwargs
+    ) -> PublishDiagnosticsParams:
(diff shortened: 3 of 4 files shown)


pyscripter:
I think you need to provide default values for Optional parameters e.g.

```python 
    def publish_diagnostics(
        self,
        params_or_uri: Union[str, PublishDiagnosticsParams],
        diagnostics: Optional[List[Diagnostic]],
        version: Optional[int],
    ):
```
should be
```python 
    def publish_diagnostics(
        self,
        params_or_uri: Union[str, PublishDiagnosticsParams],
        diagnostics: Optional[List[Diagnostic]] = None,
        version: Optional[int] = None,
    ):
```

Otherwise `server.publish_diagnostics(Params)` will fail.

alcarney:
Looks good... though I'm not sure why we need to deprecate the old signature, can't it just be extended with an optional `version` parameter?

Also don't forget your changelog entry :)

pyscripter:
> Looks good... though I'm not sure why we need to deprecate the old signature, can't it just be extended with an optional `version` parameter?

Agree

tombh:
Oh, so you both don't think we should deprecate the old signature. I'm a bit confused by that because the only argument I can think of for not starting the process of a new signature is that maintaining the current signature is just less effort.

I think there are 2 good reasons for changing.
  1. There already exists an official LSP type for precisely this purpose: [PublishDiagnosticsParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#publishDiagnosticsParams).
  2. If this type had been used to begin with, then the bug in #211 would never have occurred. In a sense, the actual issue wasn't the lack of the `version` param, but rather not using the correct type. So using the correct type gives us a separation of concerns such that the responsibility of following the LSP spec is placed on the LSP procotol module, whose core purpose is to adhere to the official spec.
 
BTW, this PR gives us _both_ signatures. So there's no breaking change yet. Users can easily get rid of the warning log _and_ benefit from automatic future compatibility, by using the new signature.

alcarney:
I suppose my resistance to moving from the old signature is more of a stylistic argument than a technical one.

While the new signature is a win for maintainability, I'd argue it's a step back for usability - with my `pygls` user hat on I know I would prefer typing 
```python
server.publish_diagnostics("file:///...", [...], version=1)
```
over 
```python
params = PublishDiagnosticsParams(uri="file:///...", diagnostics=[...], version=1)
server.publish_diagnostics(params)
```

It should be possible to update the existing signature and still make it future proof 
```python
def publish_diagnostics(self, uri: str, diagnostics: List[Diagnostic], version: Optional[int] = None, **kwargs):
    params = PublishDiagnosticsParams(
       uri=uri,
       diagnostics=diagnostics,
       version=version,
       **kwargs
    )
    self.lsp.publish_diagnostics(params) 
```
Yes, it's a little bit of extra maintenance for us to add in new fields to aid with discoverability, but at least no one is prevented from using the new fields as soon as they're available.

> Also, does anybody know why there are often 2 public methods to do the same thing like this? Therefore, we can call both `pygls.server.LanguageServer.publish_diagnostics` and `pygls.protocol.LanguageServerProtocol.publish_diagnostics`. 

Perhaps we can use this PR to decide what we want the answer to be? I'd argue the answer should be your assumption

> I assume it's because `pygls.server.LanguageServer` is the more friendly user-facing class, whilst `pygls.protocol.LanguageServerProtocol` is internal and has to be comprehensive.

Where the `LanguageServerProtocol` implements the spec as defined with methods accepting `params` objects and therefore always in line with the spec and future proof, leaving the `LanguageServer` methods free to reduce some of the boilerplate where possible.

Looking at `server.py` I can see things aren't consistent today, but we could at least decide what we want to aim for :slightly_smiling_face:

pyscripter:
> I know I would prefer typing

I only know the python jedi language server and they use server.py calls, I guess for that reason.

[My solution ](https://github.com/openlawlibrary/pygls/issues/211#issuecomment-1339546874) to this issue is along your lines, simple, and backward compatible.   

But there is a half way house which I think is preferable.   The signature at server.py stays the same (with optional version added), but the signature at protocol.py changes to PublishDiagnosticsParams.   So users have a choice to use with either.

pygls/server.py
```diff
@@ -407,9 +407,17 @@ def progress(self) -> Progress:
...
+        self,
+        uri: str,
+        diagnostics: Optional[List[Diagnostic]] = None,
+        version: Optional[int] = None,
+        **kwargs
+    ):
+        """
+        Sends diagnostic notification to the client.
+        """
+        self.lsp.publish_diagnostics(uri, diagnostics, version, **kwargs)
```
alcarney, line 420: Just a thought.

It might make sense to have this method construct and pass a `PublishDiagnosticsParams` object, so that the user doesn't see a deprecation warning on what appears to be an internal function.

But perhaps that can wait until we decide on #306

alcarney (dismissed):
Apologies for the delay, hopefully you've all had / are enjoying the holiday break!

I made a comment, but up to you on whether you want to act on it or not.

pygls/server.py
tombh: This is a really good point. I've just force-pushed this. Is it what you meant?

(Apologies for the delay too! I'm in South America at the moment so it was hooooot holiday season 🥵 hahah)

alcarney (approved):
Looks good :)

tombh:
Thanks :)

That docs PR triggered a conflict, so need a re-review 😬

alcarney (approved):

tombh:
Released in https://pypi.org/project/pygls/1.0.1

PR #304, opened 2022-12-14
fix: Correctly cast from UTF16 positions

### LINKED ISSUE #302: position_from_utf16 in workspace.py may return incorrect result.
The following code that mimics the way position_from_utf16 calculates the result, demonstrates the issue:

```python
def utf16_unit_offset(chars: str):
    """Calculate the number of characters which need two utf-16 code units.
    Arguments:
        chars (str): The string to count occurrences of utf-16 code units for.
    """
    return sum(ord(ch) > 0xFFFF for ch in chars)

s = '😋😋'

def position_from_utf16(line, pos):
    return pos - utf16_unit_offset(line[:pos])

print(position_from_utf16(s, 2))
```

The result is zero, but it should be one.

Although, this may not have a large real-world impact (you don't get many high-plane chars in python code), it would be nice to have a correct calculation.

tombh:
Fixes #302 

### DIFF
--- a/pygls/workspace.py
+++ b/pygls/workspace.py
@@ -37,13 +37,17 @@ RE_START_WORD = re.compile('[A-Za-z_0-9]*$')
 log = logging.getLogger(__name__)
 
 
+def is_char_beyond_multilingual_plane(char: str) -> bool:
+    return ord(char) > 0xFFFF
+
+
 def utf16_unit_offset(chars: str):
     """Calculate the number of characters which need two utf-16 code units.
 
     Arguments:
         chars (str): The string to count occurrences of utf-16 code units for.
     """
-    return sum(ord(ch) > 0xFFFF for ch in chars)
+    return sum(is_char_beyond_multilingual_plane(ch) for ch in chars)
 
 
 def utf16_num_units(chars: str):
@@ -59,7 +63,7 @@ def position_from_utf16(lines: List[str], position: Position) -> Position:
     """Convert the position.character from utf-16 code units to utf-32.
 
     A python application can't use the character member of `Position`
-    directly as per specification it is represented as a zero-based line and
+    directly. As per specification it is represented as a zero-based line and
     character offset based on a UTF-16 string representation.
 
     All characters whose code point exceeds the Basic Multilingual Plane are
@@ -80,14 +84,44 @@ def position_from_utf16(lines: List[str], position: Position) -> Position:
     Returns:
         The position with `character` being converted to utf-32 code units.
     """
-    try:
-        return Position(
-            line=position.line,
-            character=position.character
-            - utf16_unit_offset(lines[position.line][:position.character])
+    if len(lines) == 0:
+        return Position(0, 0)
+    if position.line >= len(lines):
+        return Position(len(lines) - 1, utf16_num_units(lines[-1]))
+
+    _line = lines[position.line]
+    _line = _line.replace('\r\n', '\n')  # TODO: it's a bit of a hack
+    _utf16_len = utf16_num_units(_line)
+    _utf32_len = len(_line)
+
+    if _utf16_len == 0:
+        return Position(position.line, 0)
+
+    _utf16_end_of_line = utf16_num_units(_line)
+    if position.character > _utf16_end_of_line:
+        position.character = _utf16_end_of_line - 1
+
+    _utf16_index = 0
+    utf32_index = 0
+    while True:
+        _is_searching_queried_position = _utf16_index < position.character
+        _is_before_end_of_line = utf32_index < _utf32_len
+        _is_searching_for_position = (
+            _is_searching_queried_position and _is_before_end_of_line
         )
-    except IndexError:
-        return Position(line=len(lines), character=0)
+        if not _is_searching_for_position:
+            break
+
+        _current_char = _line[utf32_index]
+        _is_double_width = is_char_beyond_multilingual_plane(_current_char)
+        if _is_double_width:
+            _utf16_index += 2
+        else:
+            _utf16_index += 1
+        utf32_index += 1
+
+    position = Position(line=position.line, character=utf32_index)
+    return position
 
 
 def position_to_utf16(lines: List[str], position: Position) -> Position:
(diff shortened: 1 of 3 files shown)


pygls/workspace.py
```diff
@@ -80,14 +84,42 @@ def position_from_utf16(lines: List[str], position: Position) -> Position:
...
-    except IndexError:
-        return Position(line=len(lines), character=0)
+        line = lines[-1]
+    if position.character > utf16_num_units(line) - 1:
+        log.warning(
+            f"position_from_utf16() received out of bounds index: line[{position.character}]"
+        )
+        log.debug(f"position_from_utf16() line with {utf16_num_units(line)} UTF16 units: `{line}`")
+        position.character = utf16_num_units(line) - 1
+
```
pyscripter, line 102: position can be greater than utf16_num_units(line), if for example the cursor is at the end of the line.
I  would just drop this whole block not least because utf16_num_units slows things down.
_apply_incremental_change is designed to work correctly with positions beyond the end of the line.

pygls/workspace.py
```diff
@@ -80,14 +84,42 @@ def position_from_utf16(lines: List[str], position: Position) -> Position:
...
+        return position
+    if position.line <= len(lines) - 1:
+        line = lines[position.line]
+    else:
+        log.warning(
+            f"position_from_utf16() received out of bounds index: lines[{position.line}]"
         )
-    except IndexError:
-        return Position(line=len(lines), character=0)
+        line = lines[-1]
```
pyscripter, line 95: There is no need to issue a waring.
_apply_incremental_change is designed to work correctly with positions beyond the end of the file.

I  would just replace all the above with:
```python
    if position.line >= len(lines):
        return position

    line = lines[position.line]
```

pygls/workspace.py
```diff
@@ -80,14 +84,42 @@ def position_from_utf16(lines: List[str], position: Position) -> Position:
...
+            f"position_from_utf16() received out of bounds index: line[{position.character}]"
+        )
+        log.debug(f"position_from_utf16() line with {utf16_num_units(line)} UTF16 units: `{line}`")
+        position.character = utf16_num_units(line) - 1
+
+    conventional_index = 0
+    _utf16_index = 0
+    _is_end_of_line = False
+    while True:
+        _current_char = line[conventional_index]
```
pyscripter, line 107: This would fail on empty lines

pyscripter:
Please see the comments in the code.   I would just use (skipping the comments):

```python
def position_from_utf16(lines: List[str], position: Position) -> Position:
    if position.line >= len(lines):
        # start of the line after last
        return Position(len(lines), 0)  # or return position

    line = lines[position.line]

    _utf32_len = len(line)
    _utf32_index = 0
    _utf16_index = 0
    while (_utf16_index < position.character) and (_utf32_index < _utf32_len):
        _current_char = line[line_index]
        is_double_width = is_char_beyond_multilingual_plane(_current_char)
        if (is_double_width):
            _utf16_index += 2
        else:
            _utf16_index += 1
        _utf32_index += 1

    position = Position(
        line=position.line,
        character=_utf32_index
    )
    return position

```

tombh:
Thanks for the feedback and extra test cases, they're really useful. Before making these amends, can you have a look at Pygls' [`Document.offset_at_position()`](https://github.com/openlawlibrary/pygls/blob/2a052539aef0e5ada2c619ae7df9baae5e7f8939/tests/test_document.py#L250). It's the main reason I made those conditional guards at the beginning of the function. I don't know who uses `offset_at_position()`, or what it's for, but Pygls supports it, so we should probably continue to do so.

pyscripter:
> I don't know who uses `offset_at_position()`

Well, some editors may internally store the actual text of a file instead of a list of lines like pygls does and they may wont to convert from a position to an offset to the string representation of the file.  Since offset_at_position does not account for line breaks, users of the function need to adjust for line breaks.

The function will work correctly if position_from_utf16 provides the correct result, so it should not be a concern.

tombh:
Ok I've pushed some changes. I made the variables names a bit more verbose. I added a guard clause for when a queried `Position` goes over the length of the line. And I added all your tests cases.

How does it look to you?

pygls/workspace.py
tombh: i've kept this just to satisfy this test `assert doc.offset_at_position(Position(line=1, character=5)) == 12`, which seems to make sense to me

tombh:
I'm thinking of merging this as is...

pyscripter:
_utf16_end_of_line = utf16_num_units(_line) - 1

This is definitely wrong.  See my comments above.

pygls/workspace.py
tombh: @pyscripter I need some help to understand why you consider that existing test line to be wrong. Do you disagree with this assertion `assert doc.offset_at_position(Position(line=1, character=5)) == 12`?

tombh:
Did you miss my question here? https://github.com/openlawlibrary/pygls/pull/304#discussion_r1089369599

tombh:
Just taking a small step back for a moment. This is surprisingly a lot more complicated than I first thought. We have 2 things going on here.

1. The first one, and the reason that this PR exists, is the `position_from_utf16` bug you identified in #302. And I think the PR here addresses that?
2. The second one is the tangentially associated `offset_at_position` tests that fail because of the fix for 1. (We have some comments on that in [a now stale commit](https://github.com/openlawlibrary/pygls/commit/3b36463bf1476e51758a04dfee2fc6024b388001).)

Copying your comments from point 2 here:
> But why include the control line end characters in lines?
> ...
> You also seem to assume that the control characters are always /n. This is not true on Windows or for the last line.

Very good questions. I had to look this up and you were indeed right to bring it up. If I'm interpreting it correctly the LSP spec says that that `\n` and `\r\n` should be considered and considered as the same length. [The relevant section is here](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocuments). And I've added a new test for that called `test_position_for_line_endings`.

To further complicate things, I also learnt from reading the spec that the position encodings can be negotiated by the client and server! See:
image
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#clientCapabilities

It does say that it defaults to utf-16 positioning, which is what Pygls does already. So maybe we can just create an issue to acknowledge that Pygls doesn't currently support clients requesting specific position encodings.

tombh:
I'm thinking of merging this then.

alcarney (approved):
I've never managed to get my head around this issue, so if you say it's ready I'll take your word for it! :smile:

tombh:
After 8 months, it's as ready as it'll ever be I think!

PR #352, opened 2023-07-25
Fix detection of `LanguageServer` type annotation when future annotations are used

zanieb:
## Description

The following example fails on `main` because the annotation is a string rather than a concrete type due to the use of `from __future__ import annotations`. `typing.get_type_hints` can be used to resolve the annotations for the signature to support detection in this case. 

```python
from __future__ import annotations

from pygls import server

async def example(
    foo: server.LanguageServer,
    params: None,
) -> None:
    pass


from pygls.feature_manager import has_ls_param_or_annotation

assert has_ls_param_or_annotation(example, server.LanguageServer)
```

This can also be reproduced by wrapping the annotation type in a string manually (as I do in the test in this pull request).

- ~[ ] Docstrings have been included and/or updated, as appropriate~
- ~[ ] Standalone docs have been updated accordingly~

### DIFF
--- a/pygls/feature_manager.py
+++ b/pygls/feature_manager.py
@@ -19,7 +19,7 @@ import functools
 import inspect
 import itertools
 import logging
-from typing import Any, Callable, Dict, Optional
+from typing import Any, Callable, Dict, Optional, get_type_hints
 
 from pygls.constants import (ATTR_COMMAND_TYPE, ATTR_EXECUTE_IN_THREAD, ATTR_FEATURE_TYPE,
                              ATTR_REGISTERED_NAME, ATTR_REGISTERED_TYPE, PARAM_LS)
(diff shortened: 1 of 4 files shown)


tombh:
Looks good ❤️

I'm happy to merge. But maybe @alcarney wants to give it a thumbs up too?

pygls/feature_manager.py
```diff
@@ -49,7 +49,7 @@ def has_ls_param_or_annotation(f, annotation):
     try:
         sig = inspect.signature(f)
         first_p = next(itertools.islice(sig.parameters.values(), 0, 1))
-        return first_p.name == PARAM_LS or first_p.annotation is annotation
+        return first_p.name == PARAM_LS or get_type_hints(f)[first_p.name] == annotation
```
zanieb, line 52: Oh I changed from `is` to `==` here by accident. I don't think `is` is proper though since it's not a singleton.

tombh: Yeah, I feel like `==` is most often the right choice, so it's probably a good change.

tombh:
Thank you 🙇

alcarney:
> But maybe @alcarney wants to give it a thumbs up too?

Better late than never! :sweat_smile: Looks good!

PR #354, opened 2023-07-29
chore: introduce `black` formatting

tombh:
Black is an opinionated code formatter https://black.readthedocs.io/en/stable/

This introduces a _lot_ of formatting changes.

~This has the downside of now assigning all these changes to me, the committer, which means we lose some nuance as to the history and context of the original changes. But maybe it is better to "rip the bandage" to begin enjoying consistent formatting?~ This isn't actually such a problem because of `.git-blame-ignore-revs`

TODO:
* [x] Even though Black is designed to be minimally configurable, there are some options, such as setting the max line length (defaut is 88): https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html
* [x] `scripts/generate_client.py` currently outputs non Black-compatible code. We could either make it output compatible code, or just add a compulsory Black formatting step. I think we just add the extra Black format step.


alcarney:
> This has the downside of now assigning all these changes to me, the committer, which means we lose some nuance as to the history and context of the original changes.

I stumbled across something that might help with this a little while ago - [.git-blame-ignore-revs](https://github.com/NixOS/nixpkgs/blob/master/.git-blame-ignore-revs) it looks like you can configure git to ignore some commits when doing a `git blame`

Looks like GitHub [supports it](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/) too

tombh:
Nice! Added it 😀

dimbleby:
> `scripts/generate_client.py` currently outputs non Black-compatible code

another way would be just not to format the generated code
```
[tool.black]
exclude = "pygls/lsp/whatever"
```

tombh:
Nice idea! Updated and all green CI now, thanks.

alcarney (approved):

PR #355, opened 2023-07-29
ci: retry Pyodide tests

tombh:
This is to get around a curious issue that isn't resolved by dependency locking. The flakey error is:

`ImportError: cannot import name 'Any' from 'typing_extensions'`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,9 +74,13 @@ jobs:
           sudo apt install chromium-browser chromium-chromedriver
           poetry install --with pyodide
       - name: Run Testsuite
-        run: |
-          source $VENV
-          poe test-pyodide
+        uses: nick-fields/retry@v2
+        with:
+          timeout_minutes: 10
+          max_attempts: 3
+          command: |
+            source $VENV
+            poe test-pyodide
 
   lint:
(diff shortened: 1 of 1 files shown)


alcarney:
Sorry, I keep meaning to look into this.

Looking at things more closely, I think the pip install step based on a locked set of dependencies that was added in https://github.com/openlawlibrary/pygls/commit/476eaf13c5d6e532c48b3d1f349dfcdefcfc0ca6 should probably be moved to [here](https://github.com/openlawlibrary/pygls/blob/acc239ad31b4103b79a918837ee6626bcc7ad1fc/tests/pyodide_testrunner/test-runner.js#L22)

Aside from the `selenium` stuff, `run.py` only needs to package `pygls` plus it’s test suite into a wheel so that it can be pushed inside the pyodide runtime where the tests are actually run. It’s this "hidden" pip install step inside the runtime itself that is causing the issue and needs some help picking the right version of everything.

alcarney (approved):
This looks like a great workaround for the meantime though 😄

tombh:
Ahh, yes that makes sense, it's the Pyodide runtime that needs the pinned deps. So there's still a chance 😀

But no rush, this PR should get us by.

Issue #367, opened 2023-09-01
Flakey CI errors


tombh:
## Pyodide
This might just be temporary, but for the record the error is:

```
ERROR lsp/test_code_action.py::test_code_actions - NotImplementedError
ERROR lsp/test_completion.py::test_completion - NotImplementedError
ERROR lsp/test_diagnostics.py::test_diagnostics - NotImplementedError
ERROR lsp/test_inlay_hints.py::test_code_actions - NotImplementedError
```

## Poetry Windows 3.12 installation
`D:\a\_actions\snok\install-poetry\v1/main.sh: line 36: poetry: command not found`

alcarney:
I *really* need to dig into the install step and see if we can get a reproducible set of dependencies (in the pyodide runtime!) installed.
The versions that fail are picking up a version of `pytest-asyncio` that tries to be helpful and close unclosed event loops - unfortunately closing an event loop is not implemented for `pyodide` causing the error.

The "good" runs are not picking up that version

tombh:
Oh well spotted on the cause of the error there. So I assume there's at least some version of pytest-asyncio that'll get the tests to pass? Then it's just a matter of getting the Pyodide build to use pinned deps, I seem to remember you had an idea for that?

alcarney:
> So I assume there's at least some version of pytest-asyncio that'll get the tests to pass? 

Not sure about that... it might be that it's going to be some future version... if I remember rightly the deprecation warning we see in #334 is because `pytest-asyncio` will not be able to close loops in future due to some upcoming API changes or something.

tombh:
Oh how curious! Well `pytest-asyncio` randomly showing up and disappearing in itself is an issue. I don't know how that could happen, only that `pytest`'s search path isn't deterministic? So would a fix be to simply prevent the install of `pytest-asyncio` in Pyodide?

alcarney:
> I don't know how that could happen, only that pytest's search path isn't deterministic?

It's only a hunch, but my gut tells me there's something funny going on with how the pyodide runtime installs dependencies since it uses [micropip](https://micropip.pyodide.org/en/stable/project/usage.html) rather than pip.

> So would a fix be to simply prevent the install of pytest-asyncio in Pyodide?

Probably... worst case we end up skipping some tests cases that in theory at least be run in that environment. But having *something* run reliably is probably an improvement overall?

tombh:
[...]

PR #368, opened 2023-09-01
fix: ensure server commands can be executed

### LINKED ISSUE #365: can't run language server commands in vscode-playground example
I followed the instructions to launch the vscode-playground example:

https://github.com/openlawlibrary/pygls/blob/main/examples/vscode-playground/README.md

Hitting F5 successfully launches VSCode with the example plugin loaded. I verified that the server was running by checking the output window. JSON validation is working, I get red squiggles under errors.

Now the problems start:

Selecting "pygls: Execute Command" from the command palette results in an error:

image

Furthermore, none of the commands from the json_server example are available in the command palette. For example "Count Down 10 Seconds [Non Blocking]" is not present.

I expected all of the commands that are implemented in `json_server.py` to be available and to run correctly:

https://github.com/openlawlibrary/pygls/blob/main/examples/servers/json_server.py

I tried rolling back to 05ba2c46e330fc5284d3b8d36b43846eb1844885 (just prior to the refactoring that deleted the json-vscode-extension. **EDIT**: in this case, I am able to load the json-vscode-extension example, and the commands do run. For example I am able to select  "Count Down 10 Seconds [Non Blocking]" from the command palette, and it displays the countdown notifications as expected.

alcarney:
Screenshot_20230901_191518

This brings back the ability to execute commands provided by the language server in the `vscode-playground` extension.

Instead of encoding all the commands in the `package.json` file, the extension inspects the `ServerCapabilities` of the `initialize` response to determine which commands a server provides. It then presents a quick pick list of commands the user can choose to run.

This should mean commands (that don't take any arguments at least) now "just work" with the playground extension

This PR also re-defines the example configuration keys in the `packae.json` for  `json_server.py` otherwise VSCode simply responds with `null` when the configuration values are requested

Closes #365 

### DIFF
--- a/examples/servers/json_server.py
+++ b/examples/servers/json_server.py
@@ -41,7 +41,7 @@ class JsonLanguageServer(LanguageServer):
     CMD_SHOW_CONFIGURATION_THREAD = "showConfigurationThread"
     CMD_UNREGISTER_COMPLETIONS = "unregisterCompletions"
 
-    CONFIGURATION_SECTION = "jsonServer"
+    CONFIGURATION_SECTION = "pygls.jsonServer"
 
     def __init__(self, *args):
         super().__init__(*args)
--- a/examples/vscode-playground/package.json
+++ b/examples/vscode-playground/package.json
@@ -35,6 +35,17 @@
     "configuration": [
       {
         "type": "object",
+        "title": "Json Server Configuration",
+        "properties": {
+          "pygls.jsonServer.exampleConfiguration": {
+            "scope": "resource",
+            "type": "string",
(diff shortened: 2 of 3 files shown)


alcarney:
The poetry install for windows+python 3.12 seems a little flakey too 
https://github.com/openlawlibrary/pygls/actions/runs/6052781183/job/16426922189?pr=368#step:4:67

Though since 3.12 is still a pre-release perhaps that's to be expected! :)

tombh (approved):

tombh:
Ah yes, I've added that to the #367 falkey tracking issue 😅

PR #373, opened 2023-09-07
chore(ci): bust cache

tombh:
Having problems in CI, looks like a bug got cached, hopefully just busting the cache fixes it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Use Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
+        id: setup-python
         with:
           python-version: ${{ matrix.python-version }}
           allow-prereleases: true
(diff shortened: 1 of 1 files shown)


tombh:
Actually, wait, I have another idea. Why not make `-bustme` be a Actions ENV var, so we can bust the cache just by changing the var in settings! I'll push that when I get a moment soon...

alcarney:
Also, if you know which cache is bad you should be able to delete it from here - https://github.com/openlawlibrary/pygls/actions/caches

alcarney:
I think the cache must be capturing the exact version of Python - (the [Windows error message](https://github.com/openlawlibrary/pygls/actions/runs/6122748751/job/16619306003#step:7:13) seems more useful)

>  No Python at '"C:\hostedtoolcache\windows\Python\3.11.4\x64\python.exe'

Which makes sense when you see that [earlier](https://github.com/openlawlibrary/pygls/actions/runs/6122748751/job/16619306003#step:3:10) in the run Python 3.11.5 was installed

> Successfully set up CPython (3.11.5)

Is there a way for us to get the minor Python version number into the cache key? I wonder if `actions/setup-python` exports that as a step output :thinking:

tombh:
I didn't know about that cache page! That's really good to know.

And fantastic you found the underlying cause, I added the minor version to the cache key like you suggested.

alcarney (approved):

Issue #334, opened 2023-05-08
Improve `asyncio` event loop handling


alcarney:
I've been [struggling](https://github.com/swyddfa/lsp-devtools/issues/48) to [track down](https://github.com/pytest-dev/pytest-asyncio/discussions/541) the source of this DeprecationWarning when using `pytest-lsp` with the latest `pytest-asyncio`

```
DeprecationWarning: pytest-asyncio detected an unclosed event loop when tearing down the event_loop
  fixture: <_UnixSelectorEventLoop running=False closed=False debug=False>
  pytest-asyncio will close the event loop for you, but future versions of the
  library will no longer do so. In order to ensure compatibility with future
  versions, please make sure that:
      1. Any custom "event_loop" fixture properly closes the loop after yielding it
      2. Your code does not modify the event loop in async fixtures or tests
  
    warnings.warn(
```

Long story short, I *think* I've finally tracked it down to the following block of code.

https://github.com/openlawlibrary/pygls/blob/fad812aae28bb4884b4877a5b986970941b57037/pygls/server.py#L187-L192

When giving `pygls` a specific `loop` to use, it goes ahead and creates a new one anyway, resulting in a second, unused event loop that `pytest-asyncio` is correctly warning about.
While tweaking the code above to something like 

```python
        if loop is None and IS_WIN:
            asyncio.set_event_loop(asyncio.ProactorEventLoop())
        elif loop is None and not IS_PYODIDE:
            asyncio.set_event_loop(asyncio.SelectorEventLoop())

        self.loop = loop or asyncio.new_event_loop()
```
appears to resolve the issue. I do wonder whether it's worth revisiting how `pygls` handles event loops more generally.

As far as I can tell, most of pygls' event loop code was written pre-Python 3.7 where the current [high level asyncio APIs](https://docs.python.org/3.11/library/asyncio-api-index.html) didn't exist.

I guess is this is a long winded way of asking - can anyone think of any use cases that require `pygls` to stick with the current low-level APIs? I admit, despite `pytest-lsp` forcing me to care more about event loops, I still don't fully understand them :sweat_smile:

tombh:
That's a lot of debugging! Really awesome that you're getting to the bottom of this. I think it'd be great to modernise Pygls' async code. Though of course I'm not deeply familiar with the current design decisions. The fact that you've built up so much momentum in this area now, I'm sure you'd do a great job of migrating to the high level APIs.

alcarney:
[...]

PR #375, opened 2023-09-11
UTF-8 and UTF-32 position encoding support

tombh:
This allows editors to use position encodings other than the default of UTF-16. This contributes to LSP 3.17 support. See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#positionEncodingKind


pygls/protocol.py
```diff
@@ -208,6 +214,10 @@ def default_converter():
 
     converter.register_structure_hook(JsonRPCNotification, _params_field_structure_hook)
 
+    converter.register_structure_hook(
```
tombh, line 217: @alcarney I was surprised that I needed to add this here. I feel like I must be doing something wrong or misunderstanding something? Does it make sense to you?

alcarney: I'd say it's almost expected, we're probably the first project to be exercising parts of lsprotocol's code base so it's likely we'd run into a missing structure hook here and there. (e.g. microsoft/lsprotocol#259)

Probably worth opening a PR/issue and getting it fixed upstream :)

tombh: Ohh, thanks Alex! I've added an issue over at https://github.com/microsoft/lsprotocol/issues/270

pygls/workspace/position.py
```diff
@@ -17,177 +17,175 @@
...
-        position.character = _utf16_end_of_line - 1
-
-    _utf16_index = 0
-    utf32_index = 0
-    while True:
-        _is_searching_queried_position = _utf16_index < position.character
-        _is_before_end_of_line = utf32_index < _utf32_len
-        _is_searching_for_position = (
-            _is_searching_queried_position and _is_before_end_of_line
+class Position:
```
alcarney, line 28: Another backwards compatibility note, it's probably worth adding back some top level functions for the old `utf16_unit_offset` that call out to the new `Position` class.

Perhaps, they could emit a `DeprecationWarning` so that they could be cleaned up at a later date?

alcarney (commented):
I can't comment on the actual implementation of `UTf-8` and `UTF-32` support, but overall the approach looks fine to me.

Just added some notes around maintaining backwards compatibility

karthiknadig:
@tombh I published a build of `lsprotocol` with the fix for PositionEncodingKind

tombh:
Thanks Karthik, I did see that, thank you so much. I've just been busy and haven't had a chance to update this PR yet.

Once this is merged we can make a new release with LSP 3.17 support!

tombh:
@alcarney I've pushed those changes. Also, because I'm trying to follow strict typing from now on, I made a change to the `Server`, `LanguageServer` classes, just moving the `lsp` field. I think it was to do with the converter work you did. Basically it was to better reflect that `JsonRPCProtocol` takes `LanguageServer` not `Server`. I isolated the change in the `chore: move Server.lsp to LanguageServer.lsp` commit, does that look ok?

alcarney:
> I've pushed those changes.

Thanks, they look great!

> it was to better reflect that `JsonRPCProtocol` takes `LanguageServer` not `Server`

Does it have to though? Ideally, there shouldn't be anything in `JsonRPCProtocol` that depends on `LanguageServer` directly, since we have the `LanguageServerProtocol` class to hold the details specific to LSP.

I'm sure that is not the reality though! Otherwise you wouldn't have had to update the type annotations! :smile: 
But IMO if we have to change anything we should be trying to move to a place where `JsonRPCProtocol` can be paried with `Server`, just as `LanguageServerProtocol` is paired with `LanguageServer`.

Otherwise that base `Server` class is quite useless on it's own  :sweat_smile:

----

Sort of related I've been meaning to post some thoughts in #334 on how we apply what we've learned with the `JsonRPCClient` to the server side of `pygls` (e.g. autogenerated `BaseLanguageServer` class), move to the high level asyncio APIs and generally clean up all these little inconsistencies we're finding.

It'll be mid-longer term goal though (probably resulting in a `v2`?!) . Maybe I'll try and get something written down in the next few days so we have something concrete to discuss...

tombh:
> Does it have to though? Ideally, there shouldn't be anything in JsonRPCProtocol that depends on LanguageServer directly, since we have the LanguageServerProtocol class to hold the details specific to LSP.

That totally makes sense. I've removed the commit for now and just ignored the type. I'll rejig it in the direction you mention another time. For the record these are the methods that need moving:
```
pygls/protocol.py:390: error: "Server" has no attribute "_report_server_error"  [attr-defined]
pygls/protocol.py:412: error: "Server" has no attribute "_report_server_error"  [attr-defined]
pygls/protocol.py:423: error: "Server" has no attribute "_report_server_error"  [attr-defined]
pygls/protocol.py:538: error: "Server" has no attribute "_report_server_error"  [attr-defined]
pygls/protocol.py:579: error: "Server" has no attribute "_report_server_error"  [attr-defined]
pygls/protocol.py:810: error: "Server" has no attribute "process_id"  [attr-defined]
pygls/protocol.py:812: error: "Server" has no attribute "_text_document_sync_kind"; maybe "text_document_sync_kind"?  [attr-defined]
pygls/protocol.py:813: error: "Server" has no attribute "_notebook_document_sync"  [attr-defined]
```

> Sort of related I've been meaning to post some thoughts in #334 on how we apply what we've learned with the JsonRPCClient to the server side of pygls (e.g. autogenerated BaseLanguageServer class), move to the high level asyncio APIs and generally clean up all these little inconsistencies we're finding.

How interesting! That in emulating a client you've learnt how to improve the server. Really cool.

alcarney (approved):
Looks good to me! :smile:

alcarney:
> For the record these are the methods that need moving:

Awesome thanks! They don't look too bad to fix, `_report_server_error` is generic enough that it could probably be moved into `Server`? - just with a non-LSP default reporting mechanism. 
And the others look like a type annotation on`server` [here](https://github.com/openlawlibrary/pygls/blob/f69d7eb088bf9fa3eda8c042edec8b4099d68e59/pygls/protocol.py#L732) would be enough to fix them?

tombh:
Yeah, should be pretty easy (touch wood)

PR #379, opened 2023-09-25
fix: only show code action when there's no sum

ellen364:
## Description

In examples/code_actions, tighten the ADDITION regex so that the 'Evaluate' code action only appears if there's no sum. Previously, after using the action, the lightbulb remained visible in VSCode and would keep appending the sum to the line, resulting in things like `1 + 1 = 2! 2! 2!`

### DIFF
--- a/examples/servers/code_actions.py
+++ b/examples/servers/code_actions.py
@@ -29,7 +29,7 @@ from lsprotocol.types import (
 )
 
 
-ADDITION = re.compile(r"^\s*(\d+)\s*\+\s*(\d+)\s*=\s*$")
+ADDITION = re.compile(r"^\s*(\d+)\s*\+\s*(\d+)\s*=(?=\s*$)")
 server = LanguageServer("code-action-server", "v0.1")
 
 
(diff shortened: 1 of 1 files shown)


alcarney (dismissed):

alcarney:
It looks like the regex now captures newline character at the end of the line which ending up in the code action title and failing the test. If you stripped the matched text that *should* fix it.
```diff
  action = CodeAction(
-     title=f"Evaluate '{match.group(0)}'",
+     title=f"Evaluate '{match.group(0).strip()}'",
      kind=CodeActionKind.QuickFix,
      edit=WorkspaceEdit(changes={document_uri: [text_edit]}),
  )
```

Thanks again for opening this :)

ellen364:
Thanks for all your help getting started with pygls :)

I see what you mean about the newline character. Suspect the `$` in the regex needs to be a non-capturing group, so that `match.group(0)` doesn't include the newline. Will give it a try later.

ellen364:
I've pushed a commit that changes the `ADDITION` regex to exclude trailing whitespace from the match group. All tests pass locally.

If you prefer the version that uses `strip`, I'm happy to do that instead.

PR #393, opened 2023-10-03
explicit exports from pygls.workspace

dimbleby:
so that mypy will give jedi-language-server a clean bill of health, else:
```
jedi_language_server/text_edit_utils.py:26: error: Module "pygls.workspace" does not explicitly export attribute "Workspace"  [attr-defined]
```


tombh (approved):

tombh:
Thank you ✨

PR #411, opened 2023-11-10
Remove dependency on `typeguard`

karthiknadig:
## Description (e.g. "Related to ...", etc.)

Remove dependency on `typeguard`.
fixes https://github.com/openlawlibrary/pygls/issues/409

## Code review checklist (for code reviewer to complete)

- [x] Pull request represents a single change (i.e. not fixing disparate/unrelated things in a single PR)
- [x] Title summarizes what is changing
- [x] Commit messages are meaningful (see [this][commit messages] for details)
- [x] Tests have been included and/or updated, as appropriate
- [ ] Docstrings have been included and/or updated, as appropriate
- [ ] Standalone docs have been updated accordingly

[commit messages]: https://conventionalcommits.org/

PR #374, opened 2023-09-10
Simplify vscode-playground setup and fix Python discovery

alcarney:
## Description (e.g. "Related to ...", etc.)

image

This attempts to fix the issues seen in #357 
@tombh @z80dev, would you mind trying this out and seeing if it fixes your issues?

- The multi-root setup was perhaps more complicated than necessary, so this moves `examples/workspace/` into the `examples/servers/` directory.
- The extension now launches in single-root mode  in the `examples/servers` folder by default
- The extension tries to be smarter about discovering the right interpreter to use
- If all else fails there is now a `pygls.server.pythonPath` option to override the interpreter selection made by the extension (basically how the `json-extension` used to work)
- An example `settings.json` file is now included so should be easier for the user to start tweaking values

There was also a bug in how the `json_server.py` handled the `workspace/diagnostic` request when no documents were open - that should be fixed now

## Code review checklist (for code reviewer to complete)

- [ ] Pull request represents a single change (i.e. not fixing disparate/unrelated things in a single PR)
- [ ] Title summarizes what is changing
- [ ] Commit messages are meaningful (see [this][commit messages] for details)
- [ ] Tests have been included and/or updated, as appropriate
- [ ] Docstrings have been included and/or updated, as appropriate
- [ ] Standalone docs have been updated accordingly

[commit messages]: https://conventionalcommits.org/

PR #412, opened 2023-11-16
build: v1.2.0

tombh:
* Remove dependency on `typeguard` by @karthiknadig in https://github.com/openlawlibrary/pygls/pull/411
* Simplify vscode-playground setup and fix Python discovery by @alcarney in https://github.com/openlawlibrary/pygls/pull/374
* chore: pin lsprotocol to 2023.0.0 by @alcarney in https://github.com/openlawlibrary/pygls/pull/414

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pygls"
-version = "1.1.2"
+version = "1.2.0"
 description = "A pythonic generic language server (pronounced like 'pie glass')"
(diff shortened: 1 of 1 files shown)


karthiknadig:
@tombh @alcarney `lsprotocol` 2023.0.0 (stable) is published. You can pin to that version.

I plan on publishing 2024.0.0a1 (pre-release) next week which has full LSP 3.18 and some breaking changes due to that.

alcarney (approved):
Though perhaps we should include #414  in this?

PR #417, opened 2023-11-26
feat: drop Python 3.7 support

tombh:
@karthiknadig I think it's been 3 months since Python 3.7's EOL?


pyproject.toml
```diff
@@ -13,7 +13,7 @@ license = "Apache 2.0"
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = ">=3.7.9,<4"
+python = ">=3.8.18,<4"
```
dimbleby, line 16: specifying a minor version at the lower bound is unusual - why not `>=3.8`?

(the upper bound is also controversial, at least in some circles.  Since there is no imminent prospect of python 4 it probably does not matter, but just `python = ">=3.8"` as the entire constraint should be fine).

I also notice that there are quite a few outdated dev dependencies here.  My preference for dev dependencies is `>=` constraints rather than caret or specific versions, your mileage may vary

dimbleby: also you forgot to update poetry.lock

tombh:
Great, thanks for that. How's that now? I also add a new lint sub-task to check if `poetry.lock` is up to date.

Still need to disable Python 3.7 in the Github Action matrix.

dimbleby:
looks good to me, though I think libraries should be cautious of "poetry up".  By bumping lower bounds as high as they will go, it increases the chances of unsolvable environments.  (Curiously, tight upper bounds causing the same thing seem to get a lot more attention: but an upper bound can only be incompatible with a corresponding lower bound!)

eg you have bumped websockets to ">=12": but if older versions continue to work fine for pygls then there is a case for preferring ">=11.0.3"

tests/conftest.py
```diff
@@ -85,10 +84,11 @@ def event_loop():
     of our client fixture."""
 
     # Only required for Python 3.7 on Windows.
-    if sys.version_info.minor == 7 and IS_WIN:
-        policy = asyncio.WindowsProactorEventLoopPolicy()
-    else:
-        policy = asyncio.get_event_loop_policy()
+    # @alcarney is this needed if we drop 3.7 suppoer?
+    # if sys.version_info.minor == 7 and IS_WIN:
+    #     policy = asyncio.WindowsProactorEventLoopPolicy()
```
alcarney, line 89: This should be safe to delete now

tombh: Thanks

tombh:
Thanks David, that's another really good point. What would you say is a good rule of thumb for updating dependencies? Say, for example: beyond setting a minimum version that supports required functionality, we aim to be one major version below the latest. Obviously that's not a strict rule, like for `cattrs` that doesn't follow semver, it's just a rule of thumb.

tombh:
That's fair, so at least be aware of the tradeoffs and try to find a balance without sweating too much over it.

Thanks for all your advice on this, I really appreciate it.

dimbleby (approved):

PR #420, opened 2023-11-29
ci: don't let Pyodide test fail the whole build

tombh:


alcarney (dismissed):
It looks like it's working?

alcarney:
> It still considers the whole build to have failed putting that red cross icon next to the commit message.

Strange... as the summary header in the Actions tab considered it a success :man_shrugging: 

image

> So I've just added || true for now. What do you think?

Seems fine to me

alcarney (approved):

tombh:
Yeah, I saw that too. Seems odd that they wouldn't then reflect that status in icons next to the commit messages.

PR #421, opened 2023-11-29
cattrs is a direct dependency

dimbleby:
`cattrs` is imported as a direct dependency in several places, so this project should declare that dependency.

`cattrs` uses calendar versioning so there's no reason to expect that any choice of upper bound will be safe.  I've guessed that you'll nevertheless prefer no upper bound over a pin, that seems to be the direction of travel in other issues and discussions.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1274,4 +1274,4 @@ ws = ["websockets"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.8"
-content-hash = "24c73f352815a40a121eb14a93183b1b257d33ea4cfc92908eaa303d62d36d92"
+content-hash = "5cda6a52481cb05274ed06efec69ffe66392bd08dbea32f3fe22d28609484af8"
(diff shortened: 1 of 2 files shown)


alcarney (approved):

dimbleby:
arguably this is a bit pedantic: pygls gets away without the direct dependency so long as there's an indirect dependency

but I think the implied burden of proof in your question is the wrong way round.  pygls imports and uses cattrs, why would it _not_ declare that as a requirement?

solvers are required to find a version of cattrs that is compatible both with whatever lsprotocol declares and with whatever pygls declares (and with whatever anything else in the tree declares)

alcarney:
>  pygls imports and uses cattrs, why would it not declare that as a requirement?

When you put it like that it makes perfect sense - thanks!

tombh:
@alcarney do you have ability to merge into main?

tombh:
Oh, I pipped you to the post 😬

PR #435, opened 2024-02-21
Typo: Apache License missing dash: Affects PyPI License Declaration

WilliamRoyNelson:
Poetry requires an EXACT match for the License name in order to auto-populate classifiers. Changed `Apache 2.0` to `Apache-2.0`

From version 1.1.0 forward, after switching to Poetry, PyPI has displayed the license as:
`License: Other/Proprietary License (Apache 2.0)`

PyPI **should** say: 
`License: Apache Software License (Apache 2.0)`

This is caused by the License not being an exact match to what Poetry is expecting.

This is a problem because within enterprise environments, packages are often scanned for security and legal risks. A proprietary, i.e., non-open license can be a legal issue, so it's important to have a proper license declaration to allow automated tools to make a proper evaluation.

Poetry code reference:
https://github.com/python-poetry/poetry-core/blob/219c52e6cb8b086c78f25bbfa1168d0a3096fb31/src/poetry/core/spdx/license.py#L29

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ maintainers = [
 ]
 repository = "https://github.com/openlawlibrary/pygls"
 documentation = "https://pygls.readthedocs.io/en/latest"
-license = "Apache 2.0"
+license = "Apache-2.0"
 readme = "README.md"
 
 # You may want to use the Poetry "Up" plugin to automatically update all dependencies to


tombh:
Thank you for this. Indeed it's important. There's the pedantic Conventional Commits linter that wants the first commit message to be in lower case: https://github.com/openlawlibrary/pygls/actions/runs/7994017777/job/21834287284?pr=435#step:7:27 I really should relax that requirement.

WilliamRoyNelson:
> Thank you for this. Indeed it's important. There's the pedantic Conventional Commits linter that wants the first commit message to be in lower case: https://github.com/openlawlibrary/pygls/actions/runs/7994017777/job/21834287284?pr=435#step:7:27 I really should relax that requirement.

I think I fixed the commit. The whole point of this PR is to fix a single character that's throwing off an automated tool, so it'd be a little hypocritical for me to not hold myself to the same kind of standard.

tombh (approved):

tombh:
Haha, fair comment. Great, so it's merged now. When would you like to see a formal release? I can do it soon. Otherwise I'd wait a while to see if we get any more PRs in so I can release them all at once.

WilliamRoyNelson:
I don't have an urgent need for a release. I'm sure it'll be fine if a few more PRs get merged in first.

tombh:
No worries at all. [I've just made the PR for the release](https://github.com/openlawlibrary/pygls/pull/444), as soon as it gets reviewed, I'll make the release.

tombh:
Released: https://pypi.org/project/pygls/1.3.1/

PR #458, opened 2024-05-12
Make the `pygls-playground` extension a workspace local extension

alcarney:

image

With the latest version of VSCode (`v1.89`), there is now the concept of a [workspace local](https://code.visualstudio.com/updates/v1_89#_local-workspace-extensions) extension. Any valid VSCode extension underneath the `.vscode/extensions` directory of a project are recognised by the UI and can be installed as you would any other extension.

This PR converts the `pygls-playground` extension to be workspace local which should make it a lot easier to get up and running with the example servers.

- The user still has to install dependencies and build the extension as before
- From there however, they should be able to install the extension as they would any other and use the playground directly from their main VSCode instance.

**What about older VSCode versions?**

The extension is still compatible with older versions of VSCode, using the old 2-window approach. *Assuming I've not missed anything*, the only difference should be the folder from which the 2nd debug session is launched from.


tombh:
Is it easy to try this out? I see this extension available at the moment in my VSCode 
image
It's by "noklam", is that you?

alcarney:
> I see this extension available at the moment in my VSCode

Hmm, looks like someone (not me) has published a version of the extension to the marketplace.... 

> Is it easy to try this out?

Hopefully, it's easier than it used to be! 

While you still have to build the extension yourself (steps are in the extension's README), once it's built you should be able to hit the `Install Workspace Extension` button, like in my screenshot above and have the playground Just Work<sup>TM</sup> - no need to open multiple VSCode windows anymore.

alcarney:
image

I've bumped the version number of the local extension so it should be easier to tell when you've found the right instance of it

tombh:
I'm getting stuck again I'm afraid 😞 I think it's just that I'm not familiar with VSCode. Here's some general notes:

* Would `npm install --no-save` be better so as not to modify the `package-lock.json` files?
* The instructions in `pygls-playground/README` are really good, the only thing was when it came to playing with a language server: it doesn't say that the examples are at `examples/servers/...`
* I couldn't find `Pygls > Client: Document Selector` in the command pallete.
* It wasn't immediately clear to me that `"pygls.client.documentSelector"` is in `.vscode/settings.json`.
* How are the workspace files associated with the example servers? Eg how would I know how to play with `workspace/table.txt`?
* I got an an error launching the Pygls server 🥺 `Error: spawn /home/tombh/Workspace/pygls/env/bin/python ENOENT`, see screenshot.
* I get `ECONNREFUSED` when launching the debug server.

image

alcarney:
> * Would `npm install --no-save` be better so as not to modify the `package-lock.json` files?

Sure, didn't realise that was an option! :)

> it doesn't say that the examples are at `examples/servers/...`

Good catch, I've added a note to the README.

> * I couldn't find `Pygls > Client: Document Selector` in the command pallete.
> * It wasn't immediately clear to me that `"pygls.client.documentSelector"` is in `.vscode/settings.json`.

Ah, I can see how that could be misleading, `Pygls > Client: Document Selector` was actually referring to a location in the Settings UI.
I've removed it in favour of `"pygls.client.documentSelector"` since you can't configure it via the UI anyway

> * How are the workspace files associated with the example servers? Eg
>   how would I know how to play with `workspace/table.txt`?

The only way to know is to look at the table in `examples/servers/README.md` and see which file is listed in the "Works With" column...

> * I got an an error launching the Pygls server 🥺 `Error: spawn /home/tombh/Workspace/pygls/env/bin/python ENOENT`, see screenshot.

Looks like the extension chose the wrong working directory so Python wasn't finding the server...
Rather than have the extension try to "guess", I've included it in the default options in `.vscode/settings.json`. Hopefully that approach is more robust 🤞

> * I get `ECONNREFUSED` when launching the debug server.

Probably a side effect of it choosing the wrong working directory.

tombh:
* Great updates to the README.

* > The only way to know is to look at the table in examples/servers/README.md and see which file is listed in the "Works With" column

  Oh, yes I didn't notice!

* You were right about the working directory, but your fix didn't work for me. When I manually changed it to: `"pygls.server.cwd": "/home/tombh/Workspace/pygls/examples/servers"`, everything worked 🎉

alcarney:
> You were right about the working directory, but your fix didn't work for me.

Interesting... was it completely broken? Or was the `${workspaceFolder}` "variable" expanding to the wrong dir?

tombh:
Ah right, it appear it didn't expand it:
image

alcarney:
To ask a potentially silly question :sweat_smile: .... did you restart VSCode after recompiling the extension? I don't think it's smart enough yet to detect changes to the extension after it's initially loaded.

tombh:
Ah no, so the extension needs recompiling after changing `settings.json`? Or just when the `pygls.server.cwd` field changes?

Either way, I ran `npm run compile` and restarted VSCode and still the `${workspaceFolder}` doesn't get expanded.

PR #472, opened 2024-06-07
ci: start testing against Python 3.13

alcarney:

Python 3.13 will be [coming soon](https://dev.to/hugovk/help-test-python-313-14j1).
Opening this PR to see what, if any impact 3.13 has on pygls.

We might also want to consider the [free threaded version](https://dev.to/hugovk/help-us-test-free-threaded-python-without-the-gil-1hgf) at some point?

## Automated linters

```sh
poetry run poe lint
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


alcarney:
@tombh I've now also dropped Python 3.8 from the CI matrix, it's a little early (3.8 goes end of life on [Oct 31st](https://endoflife.date/python) but since it's a breaking change I think it makes sense to do it now along with #487

If you're happy with that, it looks like we will need to also remove the 3.8 pipelines from the list of required CI jobs.

tombh (approved):

tombh:
Sounds good.

So the idea is to publish this as a release candidate along with the breaking changes in https://github.com/openlawlibrary/pygls/pull/487. That way we're not forcing dropped Python 3.8 support in our official release.

Our own CI runs will stop testing against 3.8 but that doesn't really affect the public, it just raises a very minor risk of not catching any 3.8-only bugs in the next 2 months.

I'm on board: I've disable the CI requirement for Python 3.8 in the test matrix ✅

(I've also added a new requirement for Python 3.12 tests, which I should have added a while ago.)

alcarney:
Sounds good, also gives people chance to try porting before we officially release v2

PR #516, opened 2024-12-04
Introduce support for generator feature handlers

### LINKED ISSUE #381: Dynamically register server capabilities in `initialize`
Hey there 👋 

I'm interested in dynamically registering the features that our LSP server supports, based on client configuration. Concretely, while the Ruff LSP has traditionally surfaced lint diagnostics, it can now also support code formatting, and I'd like users to be able to "turn off" the formatting and/or linting behaviors as desired, based on client configuration.

The way we solve this today is via an environment variable. The VS Code extension registers a setting in VS Code, and then the client sets an environment variable prior to starting the server. The server then [reads that variable](https://github.com/astral-sh/ruff-lsp/blob/e3da9779c887c64e5ee1eaba6da79f0f5ff46907/ruff_lsp/server.py#L65) and conditionally registers the text formatting action:

```python
if RUFF_EXPERIMENTAL_FORMATTER:

    @LSP_SERVER.feature(TEXT_DOCUMENT_FORMATTING)
    async def format_document(
        ls: server.LanguageServer,
        params: DocumentFormattingParams,
    ) -> list[TextEdit] | None:
        return await _format_document_impl(ls, params)
```

This is fine, but it won't work as well for other LSP clients. It'd be nice if this were instead driven by a configuration option that was passed from client to server...

My thinking was to instead move this into the `@LSP_SERVER.feature(INITIALIZE)` handler: so, look at the provided settings, and register functions conditionally, something like:

```python
@LSP_SERVER.feature(INITIALIZE)
def initialize(params: InitializeParams) -> None:
    """LSP handler for initialize request."""

    if params.initialization_options.get("settings", {}).get("format"):

        @LSP_SERVER.feature(TEXT_DOCUMENT_FORMATTING)
        async def format_document(
            ls: server.LanguageServer,
            params: DocumentFormattingParams,
        ) -> list[TextEdit] | None:
            return await _format_document_impl(ls, params)

    ...
```

(IIUC, this would require that the user restart the LSP when changing configuration, but that's fine.)

However, in testing, while the registration _is_ being called when I'd expect it to, VS Code is saying that the capabilities aren't being registered:

Screen Shot 2023-09-26 at 11 25 59 PM

Is this something that you would expect to work? Is there a better way to accomplish this client-driven dynamic server registration?

Thank you in advance, grateful for all the work you do on `pygls`!

### LINKED ISSUE #433: pygls does not handle async user `shutdown` handler correctly
If I register an async handler for the `shutdown` request, pygls does not guarantee that it will finish executing before responding to the client. This often leads to the client sending the `exit` notification and pygls calling `sys.exit()` before my code cleaned up everything it needs to. 

I'm finding this is leading to a lot of orphaned `esbonio` processes hanging around on my machine! :sweat_smile: 
image

I think the issue is down to how pygls handles the user registering features that "shadow" built in features.

https://github.com/openlawlibrary/pygls/blob/d3a14210c0d892bca4c2101f3f2cf9b7c843e2a3/pygls/protocol/lsp_meta.py#L15-L29

If we look at how `_execute_notification` is implemented

https://github.com/openlawlibrary/pygls/blob/d3a14210c0d892bca4c2101f3f2cf9b7c843e2a3/pygls/protocol/json_rpc.py#L144-L153

It uses `asyncio.ensure_future` to schedule the coroutine to be executed, but then will return almost immediately - meaning there's no guarantee it will complete before pygls responds to the client.

alcarney:
~This isn't quite ready to be merged just yet, but I may as well open it now for some initial feedback.~
This started out as an attempt to fix #433, but this approach should also allows us to also implement #381.

This PR unifies all handler execution under a common pattern and adds support for a new feature handler type - generator functions. 

These generators may `yield` a new `handler` function plus the `args` and `kwargs` to call it with. The yielded handler function is executed via pygls' (now standardised) execution framework meaning it can be sync, async, threaded or even another generator! Once complete the result of the yielded handler is sent back into the generator so it may continue. (`JsonRPCProtocol._run_generator`)

This PR also re-implements all of pygls' built-in feature handlers as generator functions - 
removing the need for the `LSPMeta` metaclass and associated `call_user_feature` wrapper.

While this does mean every built-in handler must explicitly `yield` to the user's handler function, the built-in handlers now get fine grained control over exactly when the user's handler is called - depending on what makes the most sense.

e.g. Before the built-in handler
```python
    @lsp_method(types.SHUTDOWN)
    def lsp_shutdown(self: JsonRPCProtocol, *args) -> None:
        """Request from client which asks server to shutdown."""

        if (user_handler := self.fm.features.get(types.SHUTDOWN)) is not None:
            yield user_handler, args, None

        # Don't cancel the future for this request!
        current_id = self.msg_id

        for msg_id, future in self._request_futures.items():
            if msg_id != current_id and not future.done():
                future.cancel()

        self._shutdown = True
        return None
```

e.g. After the built-in handler
```python
    @lsp_method(types.TEXT_DOCUMENT_DID_CHANGE)
    def lsp_text_document__did_change(self, params: types.DidChangeTextDocumentParams):
        for change in params.content_changes:
            self.workspace.update_text_document(params.text_document, change)

        if (user_handler := self.fm.features.get(types.TEXT_DOCUMENT_DID_CHANGE)) is not None:
            yield user_handler, (params,), None
```

e.g During the built-in handler
```python
    @lsp_method(types.INITIALIZE)
    def lsp_initialize(
        self, params: types.InitializeParams
    ) -> Generator[Any, Any, types.InitializeResult]:

        # Initialize the workspace before yielding to the user's initialize handler
        workspace_folders = params.workspace_folders or []
        self._workspace = Workspace(
            root_uri,
            text_document_sync_kind,
            workspace_folders,
            position_encoding,
        )

        if (user_handler := self.fm.features.get(types.INITIALIZE)) is not None:
            yield user_handler, (params,), None

        # Now that the user has had the opportunity to setup additional features, calculate
        # the server's capabilities
        self.server_capabilities = ServerCapabilitiesBuilder(
            self.client_capabilities,
            set({**self.fm.features, **self.fm.builtin_features}.keys()),
            self.fm.feature_options,
            list(self.fm.commands.keys()),
            text_document_sync_kind,
            notebook_document_sync,
            position_encoding,
        ).build()

        return types.InitializeResult(
            capabilities=self.server_capabilities,
            server_info=self.server_info,
        )

```

Other details
- For consistency, all handlers (including synchronous handlers!) are wrapped in a future
- Notifications and request handlers now use the same code path, (`JsonRPCProtocol._execute_handler`) their differences are handled by the callback passed to the future.
- If available, the current id of the request being handled is now available via a `ContextVar` (`JsonRPCProtocol.msg_id`)

Closes #381

## Automated linters

```sh
poetry run poe lint
```


alcarney:
How on Earth is it nearly the end of April already?! :sweat_smile:

tombh (approved):

tombh:
*looks at calendar* April?? I thought you opened this PR last week, well 2 weeks ago max hahaha 😅

I know I always say it, but this is such a great PR, well-documented, well-coded, well-tested, and well ... useful. I'm afraid I'm even more out of the loop than usual in terms of my ability to fully review this with the context it deserves, so if there's anything in particular you want me to look at I'm happy to pay closer attention. But from my side it's a LGTM from me 🚀

alcarney:
Thanks!

PR #520, opened 2025-03-12
fix: fallback JsonRpcException code for malformed server errors

kim-jsc:

Some LSP server implementations, do not reliably implement error handling. For instance, `gopls` may throw errors without a code when receiving malformed `textDocument/references` requests that contain invalid position data.

This behavior can cause a blocking loop in the message queue, which the user of the `pygls` library cannot catch because it occurs outside their event loop.

This patch introduces a fallback default error code in the jrpc client, ensuring that such errors propagate correctly to the user's event loop. As a result, `pygls` users can catch and handle these errors properly.

## Automated linters

```sh
poetry run poe lint
```

### DIFF
--- a/pygls/exceptions.py
+++ b/pygls/exceptions.py
@@ -29,7 +29,7 @@ class JsonRpcException(Exception):
         message = message or getattr(self.__class__, "MESSAGE")
         super().__init__(message)
         self.message = message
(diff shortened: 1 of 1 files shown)


alcarney (approved):

alcarney:
Thanks for this!

PR #526, opened 2025-05-06
Fixes error when passing max_workers

### LINKED ISSUE #525: Error when creating language server with argument 'max_workers'
When creating a `LanguageServer` with argument 'max_workers':

```
from pygls.lsp.server import LanguageServer

LanguageServer(
    name="Compliance Rules VS Code", version="<server version>", max_workers=1
)
```

An error occurs:
```
(venv) marcelldls@anon:~/wip/$ python test.py 
  File "/home/marcelldls/wip/test.py", line 3, in <module>
    LanguageServer(
  File "/home/marcelldls/wip/trash/venv/lib/python3.11/site-packages/pygls/lsp/server.py", line 77, in __init__
    super().__init__(*args, **kwargs)
  File "/home/marcelldls/wip/trash/venv/lib/python3.11/site-packages/pygls/lsp/_base_server.py", line 28, in __init__
    super().__init__(protocol_cls, converter_factory, loop, **kwargs)
TypeError: JsonRPCServer.__init__() got multiple values for argument 'max_workers'
```

I believe there is an error at: https://github.com/openlawlibrary/pygls/blob/fa0a30c2ae24e386069d2f291b741bf935ab13f6/pygls/lsp/_base_server.py#L17-L28

`loop` is not valid for `JsonRPCServer`

marcelldls:
Fixes #525

### DIFF
--- a/scripts/generate_code.py
+++ b/scripts/generate_code.py
@@ -312,10 +312,10 @@ def generate_server() -> str:
         "        self,",
         "        protocol_cls: Type[LanguageServerProtocol] = LanguageServerProtocol,",
         "        converter_factory: Callable[[], Converter] = default_converter,",
-        "        loop=None,",
+        "        max_workers: Optional[int] = None,",
         "        **kwargs,",
         "    ):",
-        "        super().__init__(protocol_cls, converter_factory, loop, **kwargs)",
+        "        super().__init__(protocol_cls, converter_factory, max_workers, **kwargs)",
         "",
         *methods,
     ]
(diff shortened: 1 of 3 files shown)


scripts/generate_code.py
```diff
@@ -312,10 +312,10 @@ def generate_server() -> str:
         "        self,",
         "        protocol_cls: Type[LanguageServerProtocol] = LanguageServerProtocol,",
         "        converter_factory: Callable[[], Converter] = default_converter,",
-        "        loop=None,",
+        "        max_workers: int | None = None,",
```
alcarney, line 315: It looks like Python 3.9 does not like this style of annotation, I expect we'll have to change it to `max_workers: Optional[int] = None` until we drop 3.9 support

marcelldls: Fair enough. Done

alcarney (approved):
Thank you!

Issue #538, opened 2025-05-29
v2 - Cache TextDocument.lines?


itsmeknt:
Apologies if I am not reading this correct, but I believe `TextDocument.lines` implemented [here](https://github.com/openlawlibrary/pygls/blob/v2.0.0a3/pygls/workspace/text_document.py#L167) will call `str.split()` at every call which will dynamically split the string every time.

If the source document is large (~10K LoC), this might incur unnecessary computation. It might not be a lot, but it may add a few milliseconds if, for example, the developer is working on an autocomplete function that needs to repeatedly reference or read `document.lines` multiple times.

Would it be reasonable to cache the result in memory somewhere so it is only called once per class instantiation?

itsmeknt:
I tried submitting a PR on this, but it seems it broke all the tests because we do indeed need to handle cases there `document.source` changes dynamically. That's why re-computation `document.source.split()` every time is necessary.

I would guess the best approach would be for the application developer (i.e. me) to handle the cache themselves outside of the Pygls library if they need it.

PR #539, opened 2025-05-29
cache document.lines after first use

itsmeknt:
## Description

Refer to [this](https://github.com/openlawlibrary/pygls/issues/538) for more info, but basically everytime `document.lines` is called, it recomputes `document.source.split()`, and I wanted to cache it to avoid the recomputation.

This is a simple in-memory cache.

I am new to this project so I'm not aware of the full life cycle of `document.source`. If it is loaded into memory only once on class instantiation and never changes (which I think is the case), then this simple in-memory cache should work. But we should allow `document.source` to be altered and modified dynamically, then this approach is ineffective since the cache won't know when it should be invalidated.

## Automated linters

```sh
poetry run poe lint
```

### DIFF
--- a/pygls/workspace/text_document.py
+++ b/pygls/workspace/text_document.py
@@ -58,6 +58,7 @@ class TextDocument(object):
 
         self._local = local
         self._source = source
+        self._lines = None
 
         self._is_sync_kind_full = sync_kind == types.TextDocumentSyncKind.Full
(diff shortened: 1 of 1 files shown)


itsmeknt:
Ah, it seems like the tests have failed, so my assumption on `document.source` being static and unchanged is incorrect! There is no value in this PR if that is the case, so please ignore.

PR #545, opened 2025-06-04
chore: add devcontainer definition and Makefile

alcarney:

Feel free to close this, but I am wondering if we'd be interested in adding a [devcontainer](https://containers.dev/) definition and some Makefile rules to the repo?

`pygls` is "just" a normal, pure Python library and so has fairly minimal requirements in terms of a development environment. But throw in the `pygls-playground` extension, pyodide test suite (and hopefully soon<sup>TM</sup>, a wasm-wasi test suite) and the list of requirements quickly starts to grow.

The provided `Makefile` aims to take you from a clean git clone to up and running in just a couple of commands. The only prerequisites being git, make, curl and a few other standard unix tools.

Some notes on the overall approach

- The devcontainer is **entirely optional**, it is there for people who like that approach and provides a known baseline against which the `Makefile` rules will just work<sup>TM</sup>.
- The `Makefile` *should* work fine outside of a container also (I use it this way 90% of the time), assuming you aren't running anything too exotic
- When bootstrapping a tool, the `Makefile` takes the following approach, (using poetry as an example)
  - If `poetry` is already available, the `Makefile` will use it
    ```make
    POETRY ?= $(shell command -v poetry)
    ```
  - Only if it can't be found will the `Makefile` attempt to install it
    ```make
    ifeq ($(strip $(POETRY)),)

    POETRY := $(BIN)/poetry

    $(POETRY): | $(UV)
	    $(UV) tool install poetry
	    $@ --version
    endif
    ```
    where `BIN = $HOME/.local/bin`

The initial version provides the following top-level targets

- `dist`: Package pygls as wheel and sdist
- `lint`: Run the lints we run in CI
- `test`: Run all tests under CPython, over stdio, tcp & websockets
- `test-pyodide`: Run the pyodide test suite
- `pygls-playground`: Setup and compile the pygls-playground extension

## Automated linters

```sh
poetry run poe lint
```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,6 @@
+FROM mcr.microsoft.com/devcontainers/base:ubuntu-24.04
+
+COPY tools.mk /
+
+RUN su vscode -c "make -f tools.mk tools" \
+    && rm tools.mk
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,30 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/ubuntu
+{
+    "name": "pygls",
+    // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+    "build": {
+	"dockerfile": "Dockerfile"
+    },
+    "containerEnv": {
+	"TZ": "UTC"
+    },
+    // Features to add to the dev container. More info: https://containers.dev/features.
+    // "features": {},
+    // Use 'forwardPorts' to make a list of ports inside the container available locally.
+    // "forwardPorts": [],
+    // Use 'postCreateCommand' to run commands after the container is created.
+    // "postCreateCommand": "uname -a",
+    // Configure tool-specific properties.
+    "customizations": {
+	"vscode": {
+	    "extensions": [
+		"charliermarsh.ruff",
+		"ms-python.python",
+	    ]
+	}
+
+    }
+    // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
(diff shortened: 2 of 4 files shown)


tombh (approved):
Absolutely useful!

I must admit I'm not a fan at all of `make`, but don't let that stop you merging. I'm curious why you say that the container stuff is optional, it seems you're saying that containers can be off-putting to some? I actually see it the other way round, whilst `Dockerfile`s certainly don't have unanimous support, I think they're more readily accepted than `Makefile`s.

But anyway that all pales into insignificance when considering what we already have for this, namely nothing 😅 So an unreserved approval from me ✅

tombh:
I've used Just as well, sure it's a bit more modern, but I don't think that'd be good here either, it's yet another syntax and thing to install. I think a plain old `sh` would be fine, which can of course be called from a `Dockerfile`. It wouldn't be guaranteed to run on Windows, but then there'd be Docker for that case.

Installing `uv` and `node` to `$HOME/.local/bin` on a _user's_ machine should be made explicit in the docs and via prompting. The preferred option is always for the user to handle installing tools. `$HOME/.local/bin` isn't guaranteed to be on `$PATH`. And even if a valid `$PATH` could be determined there isn't a guarantee that it plays nicely with all the strangeness out there like `mise`-installed tools etc. An overly dramatic example of the fallout from this is that a less experienced user never has an avenue to automatically update `uv` and `node` when some RCE vulnerability is discovered in them. They'll update their system as normal not knowing that it had no affect.

I think Docker is just the right choice here.

And sure, with a bit more work the `sh` script could be made to detect if it's running outside of Docker and prompt with instructions for the user to install `uv` and `node` themselves. I don't know of any OS that doesn't have `node` as a native package. And well the future is just `uv` isn't it, really we should just be letting it handle all Python binaries.

But like I say, what you wrote is absolutely better than what we already had! So don't go changing it. I'm just adding perspective.

PR #564, opened 2025-07-10
Fix UTF-8 position encoding

lheckemann:

`didChange` events coming from a client which prefers UTF-8 position encoding (Helix in my case) and includes multi-byte codepoints previously resulted in inconsistency between the client's view of the document and a pygls-based server's view of it as maintained by the `LanguageServerProtocol.lsp_text_document__did_change` -> `Workspace.update_text_document` -> `TextDocument.apply_change` -> `TextDocument._apply_incremental_change`.

For example, opening a document with the contents `öbc` and inserting an `a` to make it into `öabc` would result in pygls thinking the contents are `öbac`.

This PR introduces tests that show the issue, fixes them, and adjusts some incorrect tests that are broken by the fix.

## Automated linters

```sh
uv run --all-extras poe lint
```

### DIFF
--- a/pygls/workspace/position_codec.py
+++ b/pygls/workspace/position_codec.py
@@ -33,6 +33,9 @@ class PositionCodec:
     ):
         self.encoding = encoding
 
+    def __repr__(self):
+        return f"<{self.__class__.__name__}, encoding {self.encoding}>"
+
     @classmethod
     def is_char_beyond_multilingual_plane(cls, char: str) -> bool:
         return ord(char) > 0xFFFF
--- a/pygls/workspace/position_codec.py
+++ b/pygls/workspace/position_codec.py
@@ -133,18 +133,13 @@ class PositionCodec:
                 break
 
             _current_char = _line[utf32_index]
-            _is_double_width = PositionCodec.is_char_beyond_multilingual_plane(
-                _current_char
-            )
-            if _is_double_width:
-                if self.encoding == types.PositionEncodingKind.Utf32:
-                    _client_index += 1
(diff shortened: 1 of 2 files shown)


lheckemann:
I also wrote https://github.com/lheckemann/pygls/commit/48521d35bfba66369cb5531db336f30e7b1b6f57 which refactors a little harder and avoids conditionals checking the encoding type on every character, but is also a little broken. Would appreciate if a maintainer took a glance at that and said if the approach is worth pursuing, or they would prefer to stick to this less-invasive approach.

wolfskaempf:
Thanks! I can also confirm this issue from my side and would be thankful for this being merged and released.

tombh:
This looks great, I'm happy to merge as is. Thanks for the diligent work. You're other approach seems good too, although I haven't looked close enough to note the big differences. Bigger refactors are certainly welcome. I don't think the CPU cycles checking every character are a performance issue because it only iterates all the characters on a line, not on the whole document right?

Also we can merge this now so that others can benefit and then you can work on another approach in the meantime. Though bear in mind that this PR is against `main` which is currently a v2 release candidate.

lheckemann:
Alright, sounds good, thanks for the quick review! Is there a 1.x release branch I should also open it against for a backport/patch release? I don't see an immediately obvious one.

tombh:
Follow https://github.com/openlawlibrary/pygls/pull/565 for updates about the release.

tombh:
Released https://pypi.org/project/pygls/2.0.0a5 🥳

PR #583, opened 2025-10-19
docs: describe arguments unpacking for server commands in v2

perrinjerome:

With the new way of calling commands with arguments from type annotations, another significant change is that the command is now called with individual arguments and no longer a list of arguments.

The `calculate_pow` from `examples/servers/commands.py` before https://github.com/openlawlibrary/pygls/commit/87e1033fe5276e265237948aada10b77857a503b would have been something like this (it did not exist yet before this change). 

```py
@server.command("calculate.pow")
def calculate_pow(args):
    """pygls 1.x version: the list of arguments is passed as a single argument."""
    x, n = args
    logging.info("args: %r, x: %r, n: %r", args, x, n)
    return x**n
```

it is now https://github.com/openlawlibrary/pygls/blob/87e1033fe5276e265237948aada10b77857a503b/examples/servers/commands.py#L65-L69 

when called like this from client like this:

```js
await vscode.commands.executeCommand("calculate.pow", 2, 8)
```

the old version received arguments as a list ( `[2, 8]` ), see the logged message when running:

```
2025-10-19 20:17:01.877 [info] executing command: 'calculate.pow'
2025-10-19 20:17:01.885 [info] args: [2, 8], x: 2, n: 8

2025-10-19 20:17:01.885 [info] Sending data: {"id": 1, "jsonrpc": "2.0", "result": 256}
```

I believe the migration doc could be a bit more explicit regarding this change, it was not obvious for me when upgrading to pygls2 ( for reference, I missed this https://github.com/perrinjerome/vscode-zc-buildout/pull/63 )

I'm submitting here a PR to the docs to try to describe this a bit more.

## Automated linters

```sh
uv run --all-extras poe lint
```

### DIFF
--- a/docs/source/pygls/howto/migrate-to-v2.rst
+++ b/docs/source/pygls/howto/migrate-to-v2.rst
@@ -75,12 +75,14 @@ To send custom notifications in v2, use the ``notify`` method on the underlying 
 See :ref:`howto-send-custom-messages` for more details
 
 
-Server commands can now use type annotations
---------------------------------------------
+Server commands are now called with individual arguments and can use type annotations
+-------------------------------------------------------------------------------------
+
+Instead of calling sever commands with a single arguments of type list containing all the command arguments, *pygls v2* now unpacks the arguments and passes them as individual parameters to the command method.
 
 *pygls* will now inspect a function's type annotations when handling ``workspace/executeCommand`` requests, automatically converting JSON values to ``attrs`` class instances, or responding with an error if appropriate.
 
-It is **not mandatory** to start using type annotations in your command definitions, but you may notice a difference in how *pygls* calls your server command methods.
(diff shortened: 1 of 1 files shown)


tombh:
Thanks again!

alcarney:
I know it's a bit late now, but yes this looks good to me! :sweat_smile: 

We could probably do with including a few more examples from `examples/servers/commands.py` as I think changing `args` to `*args` would (almost) recover the original behaviour
```
@server.command("calculate.sum")
def calculate_sum(*args):
    """Using *args to accept any number of arguments"""
    logging.info("arguments: %r", args)
    return sum(args)
``` 

```
2025-10-25 12:04:00.170 [info] executing command: 'calculate.sum'
2025-10-25 12:04:00.174 [info] arguments: (2, 8)  # args is a tuple, not a list
Sending data: {"id": 1, "jsonrpc": "2.0", "result": 10}
```

PR #628, opened 2026-09-09
fix: ignore responses to cancelled request futures (#621)

### LINKED ISSUE #621: _handle_response raises InvalidStateError when the request future was cancelled
## Summary

`JsonRPCProtocol._handle_response` calls `future.set_result(result)` / `future.set_exception(...)` without checking whether the request future was cancelled. If the server-side task that sent a request to the client is cancelled while awaiting the response, the client's reply then crashes the message handler with `InvalidStateError`.

Observed on pygls **2.1.1** (latest); the code is unchanged on `main` (`pygls/protocol/json_rpc.py`, `_handle_response`).

## Reproduction scenario

My language server debounces an analysis task that reports progress:

1. An `async` handler task calls `await server.work_done_progress.create_async(token)` — pygls sends `window/workDoneProgress/create` to the client and registers a response future.
2. The debounce logic cancels that task (a new edit arrived) while it is awaiting the response, which cancels the pending request future.
3. The client's response arrives; `_handle_response` pops the future and calls `set_result()` on it → `InvalidStateError`.

Any server→client request awaited from a cancellable task can hit this (`workspace/configuration`, `client/registerCapability`, progress create, …).

## Traceback

```
ERROR:pygls.server:Unable to handle message
  File ".../pygls/io_.py", line 167, in run_async
    protocol.handle_message(message)
  File ".../pygls/protocol/json_rpc.py", line 513, in handle_message
    ctx.run(self._handle_response, message.id, message.result)
  File ".../pygls/protocol/json_rpc.py", line 441, in _handle_response
    future.set_result(result)
  File ".../concurrent/futures/_base.py", line 544, in set_result
    raise InvalidStateError('{}: {!r}'.format(self._state, self))
concurrent.futures._base.InvalidStateError: CANCELLED: <Future at 0x10aa2f9d0 state=cancelled>
```

(Client here is Neovim 0.12; the response to `window/workDoneProgress/create` arrived after the requesting task had been cancelled.)

## Suggested fix

Guard the response path the same way the feature-handler paths already guard cancelled futures (the v2.0.1 "prevent exception handling for cancelled futures in JsonRPCProtocol" fix covers those, but not `_handle_response`):

```python
def _handle_response(self, msg_id, result=None, error=None):
    future = self._request_futures.pop(msg_id, None)

    if not future:
        logger.warning('Received response to unknown message id "%s"', msg_id)
        return

    if future.cancelled():
        logger.debug('Received response to cancelled message "%s"', msg_id)
        return

    ...
```

Happy to send a PR if that approach looks right.

IMGillusion:
## What this fixes

Issue #621: `JsonRPCProtocol._handle_response` calls `future.set_result()` / `future.set_exception()` without checking whether the request future was already cancelled. When the task awaiting a server→client request (e.g. a debounced `workDoneProgress.create`, `workspace/configuration`) gets cancelled while awaiting the response, the client's late reply then raises `concurrent.futures.InvalidStateError` and crashes the message handler.

## Change

One guard in the shared `_handle_response`: if `future.cancelled()`, log at debug and drop the `_request_futures` bookkeeping entry without trying to settle the future. This covers both the success and error response paths, since they route through the same function.

## Tests

New `tests/test_json_rpc.py` covers: response to a cancelled future (success + error paths) no longer raises and the id is cleaned up, plus the existing unknown-id warning path. Full suite: 205 passed, 7 skipped.

Fixes #621

### DIFF
--- a/pygls/protocol/json_rpc.py
+++ b/pygls/protocol/json_rpc.py
@@ -433,6 +433,12 @@ class JsonRPCProtocol:
             logger.warning('Received response to unknown message id "%s"', msg_id)
             return
 
+        if future.cancelled():
+            logger.debug(
+                'Received response to cancelled message "%s": %s', msg_id, result
+            )
+            return
+
         if error is not None:
             logger.debug('Received error response to message "%s": %s', msg_id, error)
             future.set_exception(JsonRpcException.from_error(error))
--- a/tests/test_json_rpc.py
+++ b/tests/test_json_rpc.py
@@ -1,0 +1,65 @@
+#############################################################################
+# Copyright(c) Open Law Library. All rights reserved.                      #
+# See ThirdPartyNotices.txt in the project root for additional notices.    #
+#                                                                          #
+# Licensed under the Apache License, Version 2.0 (the "License")           #
+# you may not use this file except in compliance with the License.         #
+# You may obtain a copy of the License at                                  #
+#                                                                          #
+#     http://www.apache.org/licenses/LICENSE-2.0                           #
+#                                                                          #
+# Unless required by applicable law or agreed to in writing, software      #
+# distributed under the License is distributed on an "AS IS" BASIS,        #
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. #
+# See the License for the specific language governing permissions and      #
+# limitations under the License.                                           #
+#############################################################################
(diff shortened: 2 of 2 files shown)


tombh (approved):
Thank you. We don't have an LLM policy. But I think it's generally considered polite to indicate when a PR is generated by an agent and not the committer.

